PR #15, opened 2021-07-05
[ROAD-106] Feat: add get analysis to SnykCodeClient.

aldanchenko:
Add implementation for get analysis REST API and tests.

### DIFF
--- a/Snyk.Code.Library.Tests/SnykCode/SnykCodeClientTest.cs
+++ b/Snyk.Code.Library.Tests/SnykCode/SnykCodeClientTest.cs
@@ -2,9 +2,11 @@
 {
     using System;
     using System.Collections.Generic;
+    using System.IO;
     using System.Threading.Tasks;
     using Snyk.Code.Library.Api;
     using Snyk.Code.Library.Api.Dto;
+    using Snyk.Code.Library.Api.Dto.Analysis;
     using Snyk.Code.Library.Common;
     using Xunit;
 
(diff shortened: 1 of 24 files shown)


aldanchenko:
@michelkaporin could you please pay attention on
1. `Snyk.Code.Library.Api.Dto.Analysis` - I added additional folder for analysis DTOs because we have a lot classes. What do you think?
2. This PR must contain only code for SnykCodeClient or should we add also some AnalysisService to map results from DTO to Domain objects and additional logic?
3. In `SnykCodeClientTest` Get analysis can return WAITING status so I added additional request to get results. But don't sure is it a good idea. I don't like current logic. Maybe you have some ideas?

Thank you!

Snyk.Code.Library/Api/Dto/Analysis/AnalysisResultDto.cs
```diff
@@ -0,0 +1,38 @@
+﻿namespace Snyk.Code.Library.Api.Dto.Analysis
+{
+    /// <summary>
+    /// Code analysis result information.
+    /// </summary>
+    public class AnalysisResultDto
+    {
+        /// <summary>
+        /// Gets or sets a value indicating whether anaylysis status.
```
michelkaporin, line 9: you can omit `whether` word, otherwise it's confusing :) usually it's used for boolean properties to indicate boolean options.

Snyk.Code.Library.Tests/SnykCode/SnykCodeClientTest.cs
```diff
@@ -262,11 +409,13 @@ public async Task SnykCodeClient_ChessSessionProperApiTokenProvided_CheckPassAsy
...
 
-            _ = await snykCodeClient.LoginAsync(TestUserAgent);
+            _ = await snykCodeClient.LoginAsync("Test-VisualStudio");
 
             var status = await snykCodeClient.CheckSessionAsync();
 
             Assert.True(status.IsSucccess);
         }
+
+        private string GetResourceContent(string resourceName) => File.ReadAllText(Directory.GetCurrentDirectory() + "\\Resources\\" + resourceName);
```
michelkaporin, line 419: Use Path.Join to concatenate directories

michelkaporin:
@aldanchenko 
1. Yes, it fully makes sense. More organisation to our code structure is always for good 👍 
2. I would stick with just API communication in this PR, not to introduce domain logic here. The initial plan was to have the logic in the SnykCodeService, which can be further split into services or other additional abstractions required for working with the data (e.g. what BundleService does). Once we come to SnykCodeService implementation we should think how shall the interface look like and what do we need to implement additionally (e.g. an analysis domain object and/or AnalysisService, etc.) to make the interface work.
3. Ideally we should mock out the API and prepare our own result object (faking the network call) that the method returns (thus unit testing the behaviour of the our client) there. The logic to obtain the result and retry on wait, will go to the SnykCodeService or other abstraction we will have on top of our client, which can be covered by its own unit test. Unit tests  should be simple, test a single logic piece and not a chain of calls here. The actual chain of calls, with the real API integration shall happen in integration tests.

aldanchenko:
@michelkaporin fix by comments.

michelkaporin (approved):

PR #16, opened 2021-07-06
[ROAD-186] Feat: add mocks in BundleService for SnykCodeClient.

aldanchenko:
Now all calls for SnykCodeClient in BundleService are mocked.

### DIFF
--- a/Snyk.Code.Library/Api/Dto/BundleResponseDto.cs
+++ b/Snyk.Code.Library/Api/Dto/BundleResponseDto.cs
@@ -18,12 +18,6 @@
         }
 
         /// <summary>
-        /// Initializes a new instance of the <see cref="BundleResponseDto"/> class.
-        /// </summary>
-        /// <param name="id">Bundle id value.</param>
-        public BundleResponseDto(string id) => this.Id = id;
-
-        /// <summary>
         /// Gets or sets a value indicating whether bundle id.
         /// </summary>
         [JsonPropertyName("bundleId")]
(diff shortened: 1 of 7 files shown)


Snyk.Code.Library/Api/SnykCodeClient.cs
```diff
@@ -11,7 +11,7 @@
     /// <summary>
     /// Client for SnykCode support.
     /// </summary>
-    public class SnykCodeClient
+    public class SnykCodeClient: ISnykCodeClient
```
michelkaporin, line 14: You can inherit doc comments from the interface for the members of a class, instead of having them copy-pasted in both interface and class.
Use the following signature over the methods: 
```/// <inheritdoc />```

File: Snyk.Code.Library/Api/Dto/BundleResponseDto.cs [hunk 2]
```diff
@@ -17,6 +17,12 @@ public BundleResponseDto()
         {
         }
 
+        /// <summary>
+        /// Initializes a new instance of the <see cref="BundleResponseDto"/> class.
+        /// </summary>
+        /// <param name="id">Bundle id value.</param>
+        public BundleResponseDto(string id) => this.Id = id;
```
michelkaporin, line 24: Why do you need this constructor if all properties are public anyway? You can use 
`var x = new BundleResponseDto { Id = "dummy id" };`

michelkaporin (approved):

PR #18, opened 2021-07-13
[ROAD-162, ROAD-156] Feat: snykcode service

aldanchenko:
In this PR:
* Add SnykCodeService with one method `ScanAsync()`
* Add tests for SnykCodeService()
* Split logic to additional services - `FiltersService` and `AnalysisService`
* Add raw version of `FileCacheService` (not used in SnykCodeService right now)


aldanchenko:
**Questions:**
* Moved all service interfaces to additional Folder ("package") -> Service and service implementations to Impl folder. Is it ok?
* Added tests for `AnalysisService` and `SnykCodeService` . Maybe you could suggest other options for logic, which we need to cover?
* 

_P.S.: I can forget some of the questions I wanted to ask_

Maybe it will be better to make a call and discuss together?

Any ideas are welcome!

cc @ArtsiomCh 
cc @michelkaporin 
cc @pavel-snyk

Snyk.Code.Library/Service/Impl/SnykCodeService.cs
```diff
@@ -0,0 +1,89 @@
...
+            return await analysisService.GetAnalysisAsync(resultBundle.Id);
+        }
+
+        private IDictionary<string, string> CreateFilePathToHashDictionary(IList<string> filePaths)
+        {
+            var filePathToHashDict = new Dictionary<string, string>();
+
+            foreach (string filePath in filePaths)
+            {
+                string fileContent = File.ReadAllText(filePath, Encoding.UTF8);
```
michelkaporin, line 60: Use async whenever possible to speed up execution time.

Snyk.Code.Library/Service/Impl/SnykCodeService.cs
```diff
@@ -0,0 +1,89 @@
...
+        }
+
+        /// <inheritdoc/>
+        public async Task<AnalysisResult> ScanAsync(IList<string> filePaths)
+        {
+            var filteredFiles = await this.filtersService.FilterFilesAsync(filePaths);
+
+            var filePathToHashDict = this.CreateFilePathToHashDictionary(filteredFiles);
+
+            var bundleService = new BundleService(this.codeClient);
```
michelkaporin, line 36: Use inversion of control and provide dependency during the construction. BundleService doesn't need to be reinstantiated whenever this method is called as it doesn't hold any state.
Same applies to AnalysisService

Snyk.Code.Library/Service/Impl/AnalysisService.cs
```diff
@@ -0,0 +1,106 @@
+﻿namespace Snyk.Code.Library.Service.Impl
```
michelkaporin, line 1: Interfaces and implementations are part of the same folder in .NET unless exposed to a different project, don't introduce additional dir level here.

Snyk.Code.Library/Service/Impl/AnalysisService.cs
```diff
@@ -0,0 +1,106 @@
...
+                {
+                    System.Threading.Thread.Sleep(3000);
+                }
+
+                if (analysisResultDto.Status == AnalysisStatus.Failed)
+                {
+                    throw new SnykCodeException("SnykCode Analysis failed.");
+                }
+            }
+            while (analysisResultDto.Status != AnalysisStatus.Done);
```
michelkaporin, line 40: Prone to infinite looping, create a circuit break here by putting a time limit or adding other heuristic to determine when to stop.

Snyk.Code.Library/Service/Impl/SnykCodeService.cs
```diff
@@ -0,0 +1,89 @@
...
+            resultBundle = await bundleService.CheckBundleAsync(resultBundle.Id);
+
+            if (resultBundle.MissingFiles.Count > 0)
+            {
+                _ = await bundleService.UploadFilesAsync(resultBundle.Id, this.CreateFileHashToContentDictionary(resultBundle.MissingFiles));
+            }
+
+            var analysisService = new AnalysisService(this.codeClient);
+
+            return await analysisService.GetAnalysisAsync(resultBundle.Id);
```
michelkaporin, line 51: Don't await when returning tasks.

Snyk.Code.Library/Service/Impl/AnalysisService.cs
```diff
@@ -0,0 +1,106 @@
...
+                        Text = suggestionDto.Text,
+                        ExampleCommitDescriptions = suggestionDto.ExampleCommitDescriptions,
+                    };
+
+                    foreach (var exampleCommitFixes in suggestionDto.ExampleCommitFixes)
+                    {
+                        suggestion.Fixes.Add(new SuggestionFix
+                        {
+                            CommitURL = exampleCommitFixes.CommitURL,
+                            Line = exampleCommitFixes.Lines[0].Line,
```
michelkaporin, line 91: Does Lines array has always a single element in it?

Snyk.Code.Library/Common/ExtensionMethods.cs
```diff
@@ -0,0 +1,31 @@
+﻿namespace Snyk.Code.Library.Common
+{
+    using System.Collections;
+
+    /// <summary>
+    /// Common extension methods implementations.
+    /// </summary>
+    public static class ExtensionMethods
```
michelkaporin, line 8: Let's not introduce generic (utility) classes that can be used for everything.

You can create CollectionExtensions class for the first two methods, but I would  suggest to have `GetSha256Hash` defined not as extension, as it's not a typical case that you want to obtain Sha256 for a string.

Snyk.Code.Library/Service/Impl/SnykCodeService.cs
michelkaporin: My bad, you need it here because you await some other tasks in the method body.

aldanchenko: Thank you @michelkaporin. You dive deep in technologies 👍

Snyk.Code.Library/Service/AnalysisService.cs
```diff
@@ -0,0 +1,126 @@
...
+                    {
+                        Status = AnalysisStatus.Failed,
+                    };
+
+                    break;
+                }
+
+                counter++;
+            }
+            while (analysisResultDto.Status != AnalysisStatus.Done);
```
michelkaporin, line 59: Let's have a for-loop here, as you work with counter initialisation, condition and increment which is for-loop designed for. It's also more readable.

Snyk.Code.Library/Service/BundleService.cs
```diff
@@ -24,24 +25,56 @@ public class BundleService
...
+            var resultBundle = bundle;
+
+            while (resultBundle.MissingFiles.Count > 0)
             {
-                throw new ArgumentException("Code files list is null or empty.");
+                _ = await this.UploadFilesAsync(resultBundle.Id, this.CreateFileHashToContentDictionary(resultBundle.MissingFiles));
+
+                if (counter >= UploadFileRequestAttempts)
+                {
+                    break;
```
michelkaporin, line 58: How about we log and/or inform a user that upload has not been complete (maybe by returning some bool here)? We should avoid scenario when user gets partial results on different scans and wonders why is it the case.

Snyk.Code.Library/Service/BundleService.cs
```diff
@@ -24,24 +25,56 @@ public class BundleService
...
+        public async Task UploadMissingFilesAsync(Bundle bundle)
+        {
+            int counter = 1;
+
+            var resultBundle = bundle;
+
+            while (resultBundle.MissingFiles.Count > 0)
             {
-                throw new ArgumentException("Code files list is null or empty.");
+                _ = await this.UploadFilesAsync(resultBundle.Id, this.CreateFileHashToContentDictionary(resultBundle.MissingFiles));
```
michelkaporin, line 54: FYI you can simply write the statement without assignment, if you don't need the result :)

Snyk.Code.Library/Service/FiltersService.cs
```diff
@@ -0,0 +1,52 @@
...
+        {
+            var filters = await this.GetFiltersAsync();
+            var extensionFilters = filters.Extensions;
+            var configFileFilters = filters.ConfigFiles;
+
+            var filteredFiles = new List<string>();
+
+            foreach (string filePath in filePaths)
+            {
+                if (extensionFilters.Contains(Path.GetExtension(filePath)) || configFileFilters.Contains(Path.GetFileName(filePath)))
```
michelkaporin, line 33: `configFileFilters.Contains(Path.GetFileName(filePath)` will fail because DeepCode API returns config files begin with a slash (e.g. "/.eslintrc.js").

Snyk.Code.Library.Tests/Snyk.Code.Library.Tests.csproj
```diff
@@ -52,6 +52,9 @@
   <ItemGroup>
     <ProjectReference Include="..\Snyk.Code.Library\Snyk.Code.Library.csproj" />
   </ItemGroup>
+  <ItemGroup>
+    <Reference Include="System.Runtime.Caching" />
```
michelkaporin, line 56: Is `System.Runtime.Caching` needed here?

Snyk.Code.Library/Service/FiltersService.cs
aldanchenko: @michelkaporin could you explain a little bit more here, please? I check GetFilters API result and it's return strings without '/'.

Screenshot 2021-07-20 at 16 30 37

It seems I misunderstood something :(

michelkaporin (approved):
Would love to see handling of UploadMissingFilesAsync() when it returns false, but probably you'll do at least logging in a separate PR. Approving for now :)

aldanchenko:
what am I missing in this logic (UploadMissingFilesAsync())?

PR #20, opened 2021-07-21
[ROAD-207] Feat: add logger

aldanchenko:
Added SeriLog for extension.
Logs saved to AppData/Snyk/snyk-extension.log.

P.S.: please, note I added not so many logs following the example of how Github VS extension logging works.

### DIFF
--- a/Snyk.Code.Library/Api/SnykCodeClient.cs
+++ b/Snyk.Code.Library/Api/SnykCodeClient.cs
@@ -5,6 +5,7 @@
     using System.Net.Http;
     using System.Text;
     using System.Threading.Tasks;
+    using Serilog;
     using Snyk.Code.Library.Api.Dto;
     using Snyk.Code.Library.Api.Dto.Analysis;
     using Snyk.Common;
@@ -29,6 +30,8 @@
 
         private const string AnalysisApiUrl = "publicapi/analysis";
 
+        private static readonly ILogger Logger = LogManager.ForContext<SnykCodeClient>();
+
         private readonly HttpClient httpClient;
 
         private LoginResponseDto loginResponse;
@@ -47,6 +50,8 @@
             };
 
             this.httpClient.DefaultRequestHeaders.Add("Session-Token", token);
+
+            Logger.Debug("SnykCodeClient created with url {BaseUrl} and token {Token}.", baseUrl, token);
         }
 
         /// <summary>
@@ -66,6 +71,8 @@
         /// <returns>Analysis results with suggestions and the relative positions.</returns>
         public async Task<AnalysisResultDto> GetAnalysisAsync(string bundleId)
         {
+            Logger.Debug("Request GetAnalysis for bundle id {BundleId}.", bundleId);
+
             if (string.IsNullOrEmpty(bundleId))
             {
                 throw new ArgumentException("Bundle id is null or empty.");
@@ -97,6 +104,8 @@
         /// <returns>True if upload success.</returns>
         public async Task<bool> UploadFilesAsync(string bundleId, IEnumerable<CodeFileDto> codeFiles)
         {
+            Logger.Debug("Request UploadFiles for bundle id {BundleId}.", bundleId);
+
             if (string.IsNullOrEmpty(bundleId))
             {
                 throw new ArgumentException("Bundle id is null or empty.");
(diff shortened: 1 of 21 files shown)


Snyk.Code.Library/Api/SnykCodeClient.cs
```diff
@@ -47,6 +50,8 @@ public SnykCodeClient(string baseUrl, string token)
             };
 
             this.httpClient.DefaultRequestHeaders.Add("Session-Token", token);
+
+            Logger.Debug("SnykCodeClient created with url {BaseUrl} and token {Token}.", baseUrl, token);
```
pavel-snyk, line 54: Are we ok with logging the user token on debug level?

Snyk.Code.Library/Service/AnalysisService.cs
```diff
@@ -48,20 +54,28 @@ private async Task<AnalysisResultDto> TryGetAnalysisDtoAsync(string bundleId)
                         return analysisResultDto;
 
                     case AnalysisStatus.Failed:
+                        Logger.Warning("SnykCodeClient.GetAnalysisAsync() return Failed status. Throw SnykCodeException.");
+
```
pavel-snyk, line 58: @aldanchenko, is SnykCodeException logged if we throw it? If yes, does it make sense to do something like so?:
```
... throw new SnykCodeException("SnykCode GetAnalysisAsync() returned Failed status");
```

Maybe I'm still in JVM thinking mode :).

Snyk.Code.Library/Api/SnykCodeClient.cs
aldanchenko: Good point

Snyk.Code.Library/Service/AnalysisService.cs
aldanchenko: I added warning log because don't know if exception will be handled on upper level. But it's point. Should I remove this log?

Snyk.Common/SnykDirectory.cs
```diff
@@ -0,0 +1,24 @@
+﻿namespace Snyk.Common
+{
+    using System;
+    using System.IO;
+
+    public class SnykDirectory
```
michelkaporin, line 6: What is the plan for this class? Will it be used by Snyk Code too? If not, I would rename it to SnykCliDirectory and **not** keep it in "Snyk.Common".

Snyk.Code.Library/Snyk.Code.Library.csproj
```diff
@@ -37,6 +37,7 @@
     <Reference Include="Microsoft.Bcl.AsyncInterfaces, Version=5.0.0.0, Culture=neutral, PublicKeyToken=cc7b13ffcd2ddd51, processorArchitecture=MSIL">
       <HintPath>..\packages\Microsoft.Bcl.AsyncInterfaces.5.0.0\lib\net461\Microsoft.Bcl.AsyncInterfaces.dll</HintPath>
     </Reference>
+    <Reference Include="Serilog, Version=2.0.0.0, Culture=neutral, PublicKeyToken=24c2f752a8e58a10" />
```
michelkaporin, line 40: Don't introduce the dependency here. You have it in Common project already and you don't want to mix project responsibilities here.

Snyk.Common/SnykDirectory.cs
aldanchenko: Right now we use it in `Snyk.Common.LogManager`.

Snyk.Code.Library/Service/AnalysisService.cs
pavel-snyk: I would keep it (just without token)

Snyk.Code.Library/Snyk.Code.Library.csproj
michelkaporin: Doesn't it pick up SeriLog from Common project?

aldanchenko: No. Gives an error message :(

michelkaporin: @aldanchenko seems like package.config [lists all of transitive dependencies](https://github.com/NuGet/Home/issues/6614). We need to migrate to PackageReference format probably here. Let's address it in a separate ticket.

michelkaporin (approved):

PR #27, opened 2021-09-01
[ROAD-184] Feat: analyzers consistency

aldanchenko:
Added missing analyzers in projects.


michelkaporin (approved):
lovely job, let's continuously minimise amount of warnings as a next step :)

PR #32, opened 2021-09-15
[ROAD-295] Fix: json deserialisation error

aldanchenko:
Fix deserialisation with json like

`"{\"status\":\"ANALYZING\",\"progress\":0.5,\"complete\":false}"`.

The issue was with progress type. It was long, but progress value could be 0.5.

### DIFF
--- a/Snyk.Code.Library.Tests/SnykCode/SnykCodeClientTest.cs
+++ b/Snyk.Code.Library.Tests/SnykCode/SnykCodeClientTest.cs
@@ -416,6 +416,14 @@
             Assert.True(status.IsSucccess);
         }
 
+        [Fact]
+        public void SnykCodeClient_WaitingAnalysisResultJsonProvided_DeserialisationAnalysisJsonSuccess()
+        {
+            var analysisResultDto = Json.Deserialize<AnalysisResultDto>("{\"status\":\"ANALYZING\",\"progress\":0.5,\"complete\":false}");
+
+            Assert.NotNull(analysisResultDto);
+        }
+
         private string GetResourceContent(string resourceName) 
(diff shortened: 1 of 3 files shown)


Snyk.Code.Library.Tests/SnykCode/SnykCodeClientTest.cs
```diff
@@ -416,6 +416,14 @@ public async Task SnykCodeClient_ChessSessionProperApiTokenProvided_CheckPassAsy
             Assert.True(status.IsSucccess);
         }
 
+        [Fact]
+        public void SnykCodeClient_WaitingAnalysisResultJsonProvided_DeserialisationAnalysisJsonSuccess()
+        {
+            var analysisResultDto = Json.Deserialize<AnalysisResultDto>("{\"status\":\"ANALYZING\",\"progress\":0.5,\"complete\":false}");
```
michelkaporin, line 422: Didn't we have deserialization wrapper somewhere? We rely here on potentially different configuration of deserializer making this test to pass in future, when code between API and here won't be equal. 
If we don't have wrapper, let's create one.

Anyway, love the fact you've added a test for an existing bug!

aldanchenko: @michelkaporin Yes, this `Json` class is our wrapper right now. Please, check [here](https://github.com/snyk/snyk-visual-studio-plugin/blob/c665bb3beb608bbd00180c7f2e23e87bc5641989/Snyk.Common/Json.cs#L27).

Our you mean something other?

michelkaporin: @aldanchenko meant it indeed! Sorry, it looked too native for me 😄 nice work!

michelkaporin (approved):

PR #36, opened 2021-09-18
[ROAD-302] Fix/update snykcode params on settings update

aldanchenko:
Fix change user token and url in SnykCodeClient if user update Extension Settings.


pavel-snyk (approved):
LGTM.

PR #39, opened 2021-09-20
[ROAD-314] Feat preview release

aldanchenko:
Change version.
Fix few simple bugs.


michelkaporin:
@aldanchenko I'd love to review this PR but it seems not to be rebased. Please ping me when it's ready 🙌

aldanchenko:
@michelkaporin could you please check last changes? Thank you!

michelkaporin:
@aldanchenko sorry, but I see the same diff as in old PRs here. I see you do merges here instead of our typical git rebase workflow. Be careful, this could lead to problems when mixing both rebase and merge.

aldanchenko:
@michelkaporin I tried to fix it. How about now?

Snyk.VisualStudio.Extension/source.extension.vsixmanifest
```diff
@@ -1,16 +1,16 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
```
michelkaporin, line 1: Should we perhaps have another manifest next to the stable one with `Snyk.VisualStudio.Extension/source.extension.preview.vsixmanifest`. Such that every time we have to release preview, we can just rename the file.

Current way would mean we have to copy-paste all the information from the existing stable release which would take more time.

I'd love if we can automate all this by using PowerShell script as a post-build task. Can you look into that? Perhaps as another ticket at some point in future.

Snyk.VisualStudio.Extension/Snyk/VisualStudio/Extension/SnykExtension.cs
```diff
@@ -16,7 +16,7 @@ public class SnykExtension
         /// <summary>
         /// Integration name.
         /// </summary>
-        public const string IntegrationName = "VISUAL_STUDIO";
+        public const string IntegrationName = "VISUAL_STUDIO_PREVIEW";
```
michelkaporin, line 19: Can we pull this name from vsix the same way the version is pulled from? This will reduce amount of work to be checked off during the release process (i.e. answering the question is the integration name stable or preview for the release I am making?).

Snyk.VisualStudio.Extension/Snyk/VisualStudio/Extension/Service/SnykService.cs
```diff
@@ -161,6 +160,38 @@ public SnykApiService ApiService
             }
         }
 
+        /// <inheritdoc/>
+        public SnykApiService ApiService
```
michelkaporin, line 164: Just verifying if this code has to be here. Feel free to mark as resolved, however doesn't seem to be related to the Preview release work.

michelkaporin (changes requested):

Snyk.VisualStudio.Extension/source.extension.vsixmanifest
aldanchenko: I'll take a look

Snyk.VisualStudio.Extension/Snyk/VisualStudio/Extension/SnykExtension.cs
aldanchenko: Good point! It seems we can. But my question is: could we use it as IntegrationName. From manifest file we can get "`Snyk Vulnerability Scanner (Preview)`" string. But IntegrationName more related to IDE name. This value used for Snyk CLI `SNYK_INTEGRATION_NAME`. As I remember for `SNYK_INTEGRATION_NAME` we use something like INTELLIJ, VISUAL_STUDIO, PYTHON etc.

michelkaporin: Get method can transform manifest name into CLI-required name:
1. Drop brackets
0. Replace spaces with underscores.

michelkaporin: My bad, misinterpreted it. How about leaving it as-is as "VISUAL_STUDIO"?

aldanchenko: @michelkaporin fixed integration name :)

Snyk.VisualStudio.Extension/source.extension.vsixmanifest
michelkaporin: Should we have two manifest files next to each other for the meantime?

aldanchenko: I don't sure. This changes in dev branch, separated from main branch. As I understand you suggest to have two preview manifest files? Could you please clarify what will be in each file?

michelkaporin: > This changes in dev branch, separated from main branch. 

What happens when Snyk Code branch will be merged into master? We'll overwrite the stable version manifest, right? This poses a question how should we maintain manifest for different extension types.

Therefore, I am suggesting to have two files next to each other, namely:
- `source.extension.vsixmanifest`
- `source.extension.preview.vsixmanifest`
One will be used for stable released, another for preview release.

aldanchenko: I can make it, but I'm still don't see the need for it now. As I understand we will need second file only once - when we will move `snykcode-dev` branch to `main` branch. We could just change `source.extension.vsixmanifest` before merge to `main`. We will still need to update `source.extension.vsixmanifest` in `main`. As I understand if we will make two manifest files, we will need to support both files. It seems to me that it will only double the work. 

Please don't misunderstand me. I don't mind. I just don't understand what the plus is 🤔

michelkaporin: Let me ask you the question - how do we plan to release "Preview" versions of the extension after Snyk Code lands in `main`?

aldanchenko: Good question! Because I didn't think that we will have some other "Preview" versions 🙃

aldanchenko: Thanks @michelkaporin! Now I understand your idea! 👍 Added `source.extension.preview.vsixmanifest`

michelkaporin: 👍 Maybe there's a better way to do it than having two files but we can always explore and streamline the process in future :)

aldanchenko:
@michelkaporin thanks. Fixed the build.

.github/workflows/build.yml
```diff
@@ -1,10 +1,10 @@
-name: nightly-build 
+name: nightly-build
```
michelkaporin, line 1: why is it called nightly if we don't run it nightly?

.github/workflows/build.yml
```diff
@@ -1,10 +1,10 @@
...
 
 on: [push, pull_request]
 
 env:
   SOLUTION_FILE_PATH: .
-  DEFAULT_BRANCH: feat/nightly-build
+  DEFAULT_BRANCH: feat/snykcode-dev
```
michelkaporin, line 7: why was it different previously? 🤔

.github/workflows/build.yml
aldanchenko: This name was given at the very beginning. Maybe because it sounds beautiful :)

.github/workflows/build.yml
aldanchenko: Because we had such a brunch at the very beginning and after that it didn't changed. Although we don't use it now. 

@michelkaporin this is an important point, but maybe we won't stretch the PR review?

michelkaporin: sure, let's not stretch this PR but in that case link the ticket and we're done with this PR ✔️

aldanchenko: Added new ticket to Jira :)

michelkaporin (approved):
LGTM ✅

aldanchenko:
Thanks @michelkaporin!

PR #41, opened 2021-09-25
[ROAD-357] Fix root node item information disappear.

aldanchenko:
This PR fix issue with root node item information disappear if tool window close and open again or switch between different tool windows.


Snyk.VisualStudio.Extension/Snyk/VisualStudio/Extension/UI/Toolwindow/MessagePanel.xaml.cs
```diff
@@ -70,6 +70,11 @@ public string Text
         /// </summary>
         public void ShowScanningMessage() => this.ShowPanel(this.scanningProjectMessagePanel);
 
+        /// <summary>
+        /// Show scanning message.
+        /// </summary>
+        public void ScanningMessage() => this.ShowPanel(this.scanningProjectMessagePanel);
```
michelkaporin, line 76: Shouldn't be here as far as I understand.

PR #54, opened 2021-10-17
[ROAD-363] Feat: support filter multiple .gitignore and .dcignore

aldanchenko:
Add filter by multiple .gitignore files and multiple .dcignore files.


Snyk.Code.Library/Service/DcIgnoreService.cs
```diff
@@ -45,30 +38,71 @@ public void CreateDcIgnoreIfNeeded()
...
+                var dirFiles = projectFiles
+                    .Where(file => file.StartsWith(gitIgnoreRelativeDir))
+                    .ToList();
+
+                if (dirFiles.Count == 0)
+                {
+                    continue;
+                }
+
+                var files = this.FilterFilesByFileIgnoreList(gitIgnoreFullPath, dirFiles).ToList();
```
michelkaporin, line 93: There's no need to call `ToList()` when you only work with the interface. Checkout [this SO](https://stackoverflow.com/questions/3628425/ienumerable-vs-list-what-to-use-how-do-they-work) to get some feeling around when to use it. [This](https://medium.com/@ben.k.muller/c-ienumerable-vs-list-and-array-9f099f157f4f) seems to be a nice article covering it's ins and outs.

- In this example `.AddRange()` works fine with `IEnumerable`.
- Also there's redundant `ToList()` on :114, which doesn't provide any value.

aldanchenko: @michelkaporin fixed. Thanks! Good point! 👍

aldanchenko:
@michelkaporin thanks! Lets wait @ArtsiomCh comments 👍

ArtsiomCh:
@aldanchenko yes, seems good to merge.

PR #56, opened 2021-10-29
[ROAD-490] Fix: snykcode log issue

aldanchenko:
Now SnykCode log method will log all errors.


Snyk.Code.Library/Service/SnykCodeService.cs
```diff
@@ -37,6 +37,8 @@ public SnykCodeService(IBundleService bundleService, IAnalysisService analysisSe
         /// <inheritdoc/>
         public string GetSnykCodeErrorMessage(Exception sourceException)
         {
+            Logger.Error(sourceException, string.Empty);
```
michelkaporin, line 40: I don't think it's the right way to log error in `Get()` method. As a method caller of the method you don't expect any side effects to take place.

 My suggestion is to log the `sourceException` only if needed, i.e. in `catch()` block. Another alternative is to log the exception where it occurs (higher in the call stack).

aldanchenko: @michelkaporin for now moved log error to catch block.

michelkaporin: Let's also log a message next to the exception, it should help to find the place where an exception was caught and logged.

Smth like: `Logger.Error(sourceException, "Failed to obtain Snyk Code error message");`

Otherwise LGTM ✅

michelkaporin (approved):

Snyk.Code.Library/Service/SnykCodeService.cs
aldanchenko: @michelkaporin sorry, don't understand. You mean add message `"Failed to obtain Snyk Code error message"`? Or something other?

michelkaporin: yes

PR #58, opened 2021-11-10
feat: ✨ snykcode cache [ROAD-204]

aldanchenko:
Added cache functionality for SnykCode.

### DIFF
--- a/Snyk.Code.Library.Tests/SnykCode/CodeCacheServiceTest.cs
+++ b/Snyk.Code.Library.Tests/SnykCode/CodeCacheServiceTest.cs
@@ -1,0 +1,206 @@
+﻿namespace Snyk.Code.Library.Tests.SnykCode
+{
+    using System.Collections.Generic;
+    using System.IO;
+    using Moq;
+    using Snyk.Code.Library.Service;
+    using Snyk.Common;
+    using Xunit;
+
+    /// <summary>
+    /// Test case for <see cref="ICodeCacheService"/>.
+    /// </summary>
+    public class CodeCacheServiceTest
+    {
+        [Fact]
+        public void CodeCacheService_CodeCacheExists_RemoveDeletedFiles()
+        {
+            string projectPath = Path.Combine(Path.GetTempPath(), "CodeCacheService_CodeCacheExists_CollectDeletedFiles");
+
+            if (Directory.Exists(projectPath))
+            {
+                Directory.Delete(projectPath, true);
+            }
+
+            Directory.CreateDirectory(projectPath);
+
+            string appCsPath = Path.Combine(projectPath, "App.cs");
+
+            File.WriteAllText(appCsPath, "namespace Tests {}");
+
+            var solutionServiceMock = new Mock<ISolutionService>();
+
+            solutionServiceMock
+                .Setup(solutionService => solutionService.GetPath())
+                .Returns(projectPath);
+
+            solutionServiceMock
+                .Setup(solutionService => solutionService.GetFiles())
+                .Returns(new List<string>());
+
+            var fileProvider = new SnykCodeFileProvider(solutionServiceMock.Object);
+            var codeCacheService = new CodeCacheService(fileProvider.GetSolutionPath());
+
+            codeCacheService.Initialize(new List<string> { appCsPath });
+
+            Assert.Single(codeCacheService.GetFilePathToHashDictionary());
+
+            fileProvider.RemoveFile(appCsPath);
+
+            codeCacheService.Update(fileProvider);
+
+            Assert.Empty(codeCacheService.GetFilePathToHashDictionary());
+
+            Directory.Delete(projectPath, true);
+        }
+
+        [Fact]
+        public void CodeCacheService_CodeCacheExists_ValidCache()
+        {
+            string projectPath = Path.Combine(Path.GetTempPath(), "CodeCacheService_CodeCacheExists_ValidCache");
+
+            var solutionServiceMock = new Mock<ISolutionService>();
+
+            solutionServiceMock
+                .Setup(solutionService => solutionService.GetPath())
+                .Returns(projectPath);
+
+            var fileProvider = new SnykCodeFileProvider(solutionServiceMock.Object);
+
+            var codeCacheService = new CodeCacheService(fileProvider.GetSolutionPath());
+
+            codeCacheService.SetAnalysisResult(new Domain.Analysis.AnalysisResult());
+
+            Assert.True(codeCacheService.CacheValid());
+        }
+
+        [Fact]
+        public void CodeCacheService_CodeCacheExists_InvalidCache()
+        {
+            string projectPath = Path.Combine(Path.GetTempPath(), "CodeCacheService_CodeCacheExists_InvalidCache");
+
+            if (Directory.Exists(projectPath))
+            {
+                Directory.Delete(projectPath, true);
+            }
+
+            Directory.CreateDirectory(projectPath);
+
+            string appCsPath = Path.Combine(projectPath, "App.cs");
+
+            File.WriteAllText(appCsPath, "namespace Tests {}");
+
+            var solutionServiceMock = new Mock<ISolutionService>();
+
+            solutionServiceMock
+                .Setup(solutionService => solutionService.GetPath())
+                .Returns(projectPath);
+
+            solutionServiceMock
+                .Setup(solutionService => solutionService.GetFiles())
+                .Returns(new List<string>());
+
+            var fileProvider = new SnykCodeFileProvider(solutionServiceMock.Object);
+            var codeCacheService = new CodeCacheService(fileProvider.GetSolutionPath());
+
+            codeCacheService.SetAnalysisResult(new Domain.Analysis.AnalysisResult());
+
+            Assert.True(codeCacheService.CacheValid());
+
+            fileProvider.AddNewFile(appCsPath);
+
+            codeCacheService.Update(fileProvider);
+
+            Assert.False(codeCacheService.CacheValid());
+
+            Directory.Delete(projectPath, true);
+        }
+
+        [Fact]
+        public void CodeCacheService_UpdateFile_CacheUpdated()
+        {
+            string projectPath = Path.Combine(Path.GetTempPath(), "CodeCacheService_UpdateFile_CacheUpdated");
+
+            if (Directory.Exists(projectPath))
+            {
+                Directory.Delete(projectPath, true);
+            }
+
+            Directory.CreateDirectory(projectPath);
+
+            string appCsPath = Path.Combine(projectPath, "App.cs");
+
+            File.WriteAllText(appCsPath, "namespace Tests {}");
+
+            var solutionServiceMock = new Mock<ISolutionService>();
+            var filtersServiceMock = new Mock<IFiltersService>();
+
+            solutionServiceMock
+                .Setup(solutionService => solutionService.GetPath())
+                .Returns(projectPath);
+
+            solutionServiceMock
+                .Setup(solutionService => solutionService.GetFiles())
+                .Returns(new List<string>());
+
+            var fileProvider = new SnykCodeFileProvider(solutionServiceMock.Object);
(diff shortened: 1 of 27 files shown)


Snyk.Code.Library/Service/CodeCacheService.cs
```diff
@@ -0,0 +1,213 @@
...
+    using Snyk.Common;
+
+    /// <inheritdoc/>
+    public class CodeCacheService : ICodeCacheService
+    {
+        private static readonly ILogger Logger = LogManager.ForContext<CodeCacheService>();
+
+        private ObjectCache filePathToHashCache = new MemoryCache("FilePathToFileHashCache");
+
+        private ObjectCache filePathToContentCache = new MemoryCache("FilePathToFileContentCache");
```
michelkaporin, line 19: Do we need to provide expiration time for cache? E.g. 24 hours after which we'd do a re-scan?

Snyk.Code.Library/Service/CodeCacheService.cs
```diff
@@ -0,0 +1,213 @@
...
+                {
+                    filePathToHashDict.Add(filePath, this.filePathToHashCache[filePath].ToString());
+                }
+            }
+
+            return filePathToHashDict;
+        }
+
+        /// <inheritdoc/>
+        public bool CacheNotExists() => this.analysisResult == null;
```
michelkaporin, line 96: Let's not have negation in method names and invert the boolean for this method callers. Otherwise, it's not readable.

Snyk.Code.Library/Service/ICodeCacheService.cs
```diff
@@ -0,0 +1,100 @@
...
+        /// </summary>
+        /// <param name="files">Files to initialize cache.</param>
+        void Initialize(IEnumerable<string> files);
+
+        /// <summary>
+        /// Check is cache valid for this project.
+        /// </summary>
+        /// <param name="project">Project key.</param>
+        /// <returns>True if cache for this project exists and valid.</returns>
+        bool CacheValid();
```
michelkaporin, line 53: .NET community usually prefixes boolean return methods with "Is". Let's change this to `IsCacheValid()`.

Snyk.Code.Library/Service/ICodeCacheService.cs
```diff
@@ -0,0 +1,100 @@
...
+        /// </summary>
+        /// <param name="fileProvider">File provider instance.</param>
+        void Update(IFileProvider fileProvider);
+
+        /// <summary>
+        /// Get file paths in relative format. Input will be like C:\Test\ProjectName\somefile.txt and output will be like /somefile.txt.
+        /// </summary>
+        /// <param name="files">List of absolute file paths.</param>
+        /// <returns>List of relative file paths.</returns>
+        IEnumerable<string> GetRelativeFilePaths(IEnumerable<string> files);
```
michelkaporin, line 98: How this is related to cache service work? This looks more like a helper method or something that must be defined outside of the service. Let's not violate single responsibility principle.

Snyk.Code.Library/Service/SnykCodeService.cs
```diff
@@ -57,15 +73,79 @@ public async Task<AnalysisResult> ScanAsync(IFileProvider fileProvider, Cancella
...
+        }
+
+        private void UpdateCache(string bundleId, AnalysisResult analysisResult)
+        {
+            this.codeCacheService.SetAnalysisResult(analysisResult);
+
+            this.codeCacheService.SetCachedBundleId(bundleId);
+        }
+
+        private bool NoChangedFilesInSolution(IEnumerable<string> files) => files.IsNullOrEmpty();
```
michelkaporin, line 149: I like the expressiveness here 👍 However, I'd invert method name negation and move negation to the caller.

Snyk.Code.Library/Service/SnykCodeService.cs
```diff
@@ -57,15 +73,79 @@ public async Task<AnalysisResult> ScanAsync(IFileProvider fileProvider, Cancella
...
+            fileProvider.ClearHistory();
+
+            return analysisResult;
         }
+
+        private async Task<AnalysisResult> UpdatePreviousScanAsync(IFileProvider fileProvider, IEnumerable<string> changedFiles, CancellationToken cancellationToken = default)
+        {
+            this.codeCacheService.Update(fileProvider);
+
+            IDictionary<string, string> extendFilePathToHashDict = this.codeCacheService.GetFilePathToHashDictionary(changedFiles);
```
michelkaporin, line 116: please not type variables out, because it's hard to read. Method name is already clear on what the return type is.

Snyk.Code.Library/Service/SnykCodeFileProvider.cs
```diff
@@ -17,111 +13,104 @@ public class SnykCodeFileProvider : IFileProvider
     {
         private static readonly ILogger Logger = LogManager.ForContext<SnykCodeFileProvider>();
 
-        private ObjectCache filePathToHashCache = new MemoryCache("FilePathToFileHashCache");
+        private IList<string> removedFiles;
 
-        private ObjectCache filePathToContentCache = new MemoryCache("FilePathToFileContentCache");
+        private IList<string> changedFiles;
+
+        private IList<string> addedFiles;
```
michelkaporin, line 20: Why do you need to track removed, added and changed files separately? In VSCE we count both removed and added as changed, extending bundle with them. see https://github.com/snyk/vscode-extension/blob/8f990f0f82af30e65ecb73f01385ad030685c6fe/src/snyk/snykCode/watchers/filesWatcher.ts#L19

There's no difference to the way we interact with API between removed/changed/added, is there?

Snyk.Code.Library/Service/SnykCodeFileProvider.cs
```diff
@@ -17,111 +13,104 @@ public class SnykCodeFileProvider : IFileProvider
...
-        }
+        /// <inheritdoc/>
+        public IEnumerable<string> GetAllChangedFiles() => this.GetAddedFiles()
+                .Concat(this.GetChangedFiles())
+                .Concat(this.GetRemovedFiles())
+                .Distinct()
+                .ToList();
 
-        private string CreateRelativeFilePath(string filePath)
+        private IList<string> GetDirectoryFiles()
```
michelkaporin, line 106: Why not to use `GetSolutionDirectoryFiles()` from the SolutionService instead?

Snyk.Code.Library/Service/CodeCacheService.cs
```diff
@@ -0,0 +1,213 @@
...
+
+                this.AddToFilePathToHashCache(relativeFilePath, fileHash);
+
+                this.AddToFilePathToContentCache(relativeFilePath, fileContent);
+
+                this.Invalidate();
+            }
+            catch (Exception ex)
+            {
+                Logger.Error(ex.Message);
```
michelkaporin, line 176: Let's make it clearer where the exception was caught and add messages to the log, e.g. `Failed to update cache. {ex.Message}`

Snyk.Code.Library/Service/CodeCacheService.cs
```diff
@@ -0,0 +1,213 @@
...
+        }
+
+        /// <inheritdoc/>
+        public string GetCachedBundleId() => this.bundleId;
+
+        /// <inheritdoc/>
+        public void SetCachedBundleId(string id) => this.bundleId = id;
+
+        /// <inheritdoc/>
+        public void Update(IFileProvider fileProvider)
```
michelkaporin, line 128: I don't think you need to differentiate between added/removed/changed. Changed is enough.

Snyk.Code.Library.Tests/SnykCode/CodeCacheServiceTest.cs
```diff
@@ -0,0 +1,206 @@
...
+
+            Assert.True(codeCacheService.CacheValid());
+        }
+
+        [Fact]
+        public void CodeCacheService_CodeCacheExists_InvalidCache()
+        {
+            string projectPath = Path.Combine(Path.GetTempPath(), "CodeCacheService_CodeCacheExists_InvalidCache");
+
+            if (Directory.Exists(projectPath))
```
michelkaporin, line 82: Why not to add the code that repeats from test to test to the setup "hook"? See https://xunit.net/docs/shared-context.

Snyk.Code.Library/Service/CodeCacheService.cs
aldanchenko: Could you please clarify? Maybe some example?

aldanchenko: You mean IsCacheExists?

Snyk.Code.Library/Service/SnykCodeService.cs
aldanchenko: Something like IsFilesChangedInSolution()?

Snyk.Code.Library/Service/SnykCodeFileProvider.cs
aldanchenko: @michelkaporin hm.. interesting, for VS I did it because this:
https://github.com/snyk/snyk-visual-studio-plugin/blob/cb3f8f9ccff37610b4273cd6d188641eb2a76d2d/Snyk.Code.Library/Api/SnykCodeClient.cs#L134

Snyk.Code.Library/Service/SnykCodeFileProvider.cs
aldanchenko: If I understand correctly: VS Solution could return empty files list even if solution have files. So in case VS solution instance will return empty list it will just get all solution directory files.

Snyk.Code.Library/Service/CodeCacheService.cs
michelkaporin: yep, CacheExists() or IsCacheExists() 👍

Snyk.Code.Library/Service/SnykCodeFileProvider.cs
michelkaporin: Do we track those in JB too?

Snyk.Code.Library/Service/SnykCodeFileProvider.cs
michelkaporin: My question is rather why not to reuse the method from SnykSolutionService? The code here is duplicated in two files.

Snyk.Code.Library/Service/SnykCodeFileProvider.cs
aldanchenko: As I understand from JB, yes, it's use removed files list. Also ask in Slack about this, to know for sure

Snyk.Code.Library/Service/CodeCacheService.cs
```diff
@@ -12,15 +12,17 @@
     /// <inheritdoc/>
     public class CodeCacheService : ICodeCacheService
     {
+        private const double ExprirationTime = 24.0;
```
michelkaporin, line 15: ExpirationTime

Snyk.Code.Library/Service/SnykCodeService.cs
michelkaporin: Yes, or `AnyFilesChangedInSolution` to be grammatically sound :)

Snyk.Code.Library.Tests/SnykCode/SnykCodeServiceTest.cs
```diff
@@ -8,21 +8,220 @@
...
+                    },
+                    new FileAnalysis
+                    {
+                        FileName = "app2.js",
+                    },
+                },
+            };
+
+            var filtersServiceMock = new Mock<IFiltersService>();
+            var bundleServiceMock = new Mock<IBundleService>();
```
michelkaporin, line 166: Can we have common mocks initialisation between different test cases in the constructor? Current method body has some redundancy and has many constructor calls.

michelkaporin (approved):
Let's address all the comments and merge this :) 

Amazing work here @aldanchenko 👏

PR #60, opened 2021-11-16
fix 🐛 stop scan button [ROAD-434]

aldanchenko:
Fix stop scan thread.
Fix buttons state.

### DIFF
--- a/Snyk.VisualStudio.Extension/Snyk/VisualStudio/Extension/Service/SnykTasksService.cs
+++ b/Snyk.VisualStudio.Extension/Snyk/VisualStudio/Extension/Service/SnykTasksService.cs
@@ -24,17 +24,11 @@
 
         private static SnykTasksService instance;
 
-        private CancellationTokenSource cliScanTokenSource;
+        private CancellationTokenSource ossScanTokenSource;
 
         private CancellationTokenSource snykCodeScanTokenSource;
 
         private CancellationTokenSource downloadCliTokenSource;
-
-        private Task ossScanTask;
-
-        private Task snykCodeScanTask;
-
-        private Task downloadCliTask;
 
         private ISnykServiceProvider serviceProvider;
 
@@ -160,11 +154,11 @@
         /// </summary>
         public void CancelTasks()
         {
-            if (this.cliScanTokenSource != null)
+            if (this.ossScanTokenSource != null)
             {
                 Logger.Information("Cancel OSS task");
 
-                this.cliScanTokenSource.Cancel();
+                this.ossScanTokenSource.Cancel();
 
                 this.cli?.ConsoleRunner?.Stop();
             }
@@ -198,6 +192,15 @@
         public async Task ScanAsync()
         {
             Logger.Information("Enter Scan method");
+
+            if (!this.serviceProvider.SolutionService.IsSolutionOpen)
+            {
+                this.FireOssError("No open solution");
+
+                Logger.Information("Solution not opened");
+
+                return;
+            }
 
             var selectedFeatures = await this.GetFeaturesSettingsAsync();
 
@@ -233,7 +236,7 @@
 
             Logger.Information("Start run task");
 
-            this.downloadCliTask = Task.Run(
+            _ = Task.Run(
                 () =>
                 {
                     try
@@ -266,9 +269,13 @@
                     }
                     catch (Exception exception)
                     {
-                        Logger.Information(exception.Message);
+                        Logger.Error(exception.Message);
 
                         this.OnDownloadCancelled(exception.Message);
+                    }
+                    finally
+                    {
+                        this.downloadCliTokenSource.Dispose();
                     }
                 }, progressWorker.TokenSource.Token);
         }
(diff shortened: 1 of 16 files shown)


Snyk.VisualStudio.Extension/Snyk/VisualStudio/Extension/Service/SnykTasksService.cs
```diff
@@ -263,10 +257,14 @@ public void Download(CliDownloadFinishedCallback downloadFinishedCallback = null
                             lastCliReleaseDate,
                             progressWorker: progressWorker,
                             downloadFinishedCallbacks: downloadFinishedCallbacks);
+
+                        this.downloadCliTokenSource = null;
```
michelkaporin, line 261: I think Dispose() would be the right method to call here on the token source. You can unify two statements into one in the `finally {}` clause :)

Snyk.VisualStudio.Extension/Snyk/VisualStudio/Extension/Service/SnykTasksService.cs
```diff
@@ -438,11 +439,12 @@ private void ScanOss(FeaturesSettings featuresSettings)
                         this.FireScanningCancelledEvent();
 
                         this.cli = null;
+                        this.ossScanTokenSource = null;
```
michelkaporin, line 442: `Dipose()` ?

Snyk.VisualStudio.Extension/Snyk/VisualStudio/Extension/Service/SnykTasksService.cs
```diff
@@ -465,63 +467,84 @@ private void ScanSnykCode(FeaturesSettings featuresSettings)
...
+            Logger.Information("Start scan task");
+
+            this.FireSnykCodeScanningStartedEvent();
+
+            try
             {
-                TasksService = this,
-                TokenSource = this.snykCodeScanTokenSource,
-            };
+                if (!this.serviceProvider.SolutionService.IsSolutionOpen)
```
michelkaporin, line 476: This check has to happen somewhere in a higher level, validating both OSS and Snyk Code scan. It doesn't make sense to have this check only for Snyk Code but not for OSS, does it?

Snyk.VisualStudio.Extension/Snyk/VisualStudio/Extension/Service/SnykTasksService.cs
```diff
@@ -465,63 +467,84 @@ private void ScanSnykCode(FeaturesSettings featuresSettings)
...
+                    var filesProvider = this.serviceProvider.SolutionService.NewFileProvider();
 
-                    try
+                    var analysisResult = await this.serviceProvider.SnykCodeService.ScanAsync(filesProvider, this.snykCodeScanTokenSource.Token);
+
+                    this.FireScanningUpdateEvent(analysisResult);
+
+                    this.FireSnykCodeScanningFinishedEvent();
+
+                    this.snykCodeScanTokenSource = null;
```
michelkaporin, line 495: Let's dispose and move to finally

Snyk.VisualStudio.Extension/Snyk/VisualStudio/Extension/Service/SnykTasksService.cs
```diff
@@ -465,63 +467,84 @@ private void ScanSnykCode(FeaturesSettings featuresSettings)
...
-                            {
-                                var filesProvider = this.serviceProvider.SolutionService.NewFileProvider();
+                this.FireScanningCancelledEvent();
+            }
+        }
 
-                                var analysisResult = await this.serviceProvider.SnykCodeService.ScanAsync(filesProvider, progressWorker.TokenSource.Token);
+        private bool IsTaskCancelled(Exception e)
+        {
+            var exceptionType = e.GetType();
```
michelkaporin, line 525: How does this deal with exceptions are of AggregateTypeException type with TaskCanceledException inside?

aldanchenko: Sorry @michelkaporin, maybe I misunderstood your question. As I understand it's not AggregateTypeException. There could be two different exception types, but both contains additional property CancellationToken.

Snyk.VisualStudio.Extension/Snyk/VisualStudio/Extension/Service/SnykTasksService.cs
aldanchenko: Thanks, @michelkaporin! I totally forgot about finally block :(

Snyk.VisualStudio.Extension/Snyk/VisualStudio/Extension/Service/SnykTasksService.cs
michelkaporin: I might be wrong but it could be that if we parallelise some of the work (e.g. for CLI scan) down the code, then `TaskCanceledException` will be thrown in multiple instances of tasks. This would lead to .NET wrapping it in so-called "AggregateException" and bubbling it up to here. In that case we have to inspect InnerExceptions and see if one of them was "TaskCanceledException" or "OperationCanceledException".

Even if it's not the case now that multiple tasks could throw several exceptions, it's always better to be on safe side as we don't know how the code will develop in future.

See https://docs.microsoft.com/en-us/dotnet/standard/parallel-programming/exception-handling-task-parallel-library for more understanding what AggregateException essentially is.

aldanchenko: @michelkaporin something like now will be good?

michelkaporin: @aldanchenko let's resolve branch conflicts and I'll take a look 👌

Snyk.VisualStudio.Extension/Snyk/VisualStudio/Extension/Service/SnykTasksService.cs
```diff
@@ -473,55 +474,84 @@ private void ScanSnykCode(FeaturesSettings featuresSettings)
...
+                if (canceledException.CancellationToken.IsCancellationRequested)
+                {
+                    return true;
+                }
+            }
 
-                                this.OnSnykCodeError(errorMessage);
-                            }
-                        });
+            if (exceptionType == typeof(OperationCanceledException))
```
michelkaporin, line 544: `TaskCanceledException` inherits from `OperationCanceledException`. That means you can simplify this code to also use `is`-operator, see https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/operators/type-testing-and-cast#is-operator

michelkaporin (approved):
I've left one more comment, otherwise looks good to me! Great work here 👏 hope it solves the original issue and boost the perf.

PR #61, opened 2021-11-22
Feat: ✨ support new code api endpoints

aldanchenko:
Replace old SnykCode API endpoints with new one.
Bundle payload without files property.
Removed /file endpoint.
Changed get analysis payload.
Files content uploaded via extend bundle endpoint.

### DIFF
--- a/Snyk.Code.Library.Tests/SnykCode/BundleServiceTest.cs
+++ b/Snyk.Code.Library.Tests/SnykCode/BundleServiceTest.cs
@@ -33,7 +33,7 @@
 
             var bundleDto = new BundleResponseDto
             {
-                Id = bundle.Id,
+                Hash = bundle.Id,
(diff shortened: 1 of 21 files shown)


Snyk.Code.Library.Tests/SnykCode/BundleServiceTest.cs
```diff
@@ -32,7 +32,7 @@ public async Task BundleService_TwoFilesProvided_UploadSuccessfullAfterTreeAttem
 
             var bundleDto = new BundleResponseDto
             {
-                Id = bundle.Id,
```
bastiandoetsch, line 35: Would it make sense to rename `Id` to `hash`?

bastiandoetsch (commented):
LGTM. Love the extensive testing 👏🏻

bastiandoetsch (approved):

PR #62, opened 2021-11-23
Fix 🐛 stop scan button [ROAD-434]

aldanchenko:
Fix exception on dispose call.

### DIFF
--- a/Snyk.Code.Library/Service/SnykCodeFileProvider.cs
+++ b/Snyk.Code.Library/Service/SnykCodeFileProvider.cs
@@ -50,7 +50,9 @@
         public void RemoveFile(string file) => this.removedFiles.Add(file);
 
         /// <inheritdoc/>
-        public IEnumerable<string> GetChangedFiles() => this.changedFiles.Except(this.removedFiles);
+        public IEnumerable<string> GetChangedFiles() => this.changedFiles
+            .Except(this.removedFiles)
+            .Where(file => file != null);
 
         /// <inheritdoc/>
         public IEnumerable<string> GetRemovedFiles() => this.removedFiles;
@@ -66,6 +68,6 @@
         public IEnumerable<string> GetAllChangedFiles() => this.GetChangedFiles()
                 .Concat(this.GetRemovedFiles())
                 .Distinct()
-                .ToList();
+                .Where(file => file != null);
     }
 }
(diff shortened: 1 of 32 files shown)


Snyk.VisualStudio.Extension/Snyk/VisualStudio/Extension/Service/SnykTasksService.cs
```diff
@@ -626,5 +612,45 @@ private IList<string> GetSelectedFeatures(FeaturesSettings featuresSettings)
...
+            {
+                tokenSource = null;
+            }
+        }
+
+        private void DisposeCancellationTokenSource(ref CancellationTokenSource tokenSource)
+        {
+            try
+            {
+                if (tokenSource != null)
```
michelkaporin, line 641: You can write it also like `tokenSource?.Dispose()`. See https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/operators/member-access-operators#null-conditional-operators--and-

Snyk.VisualStudio.Extension/Snyk/VisualStudio/Extension/Service/SnykTasksService.cs
```diff
@@ -626,5 +612,45 @@ private IList<string> GetSelectedFeatures(FeaturesSettings featuresSettings)
 
             return selectedProducts;
         }
+
+        private void CancelTask(ref CancellationTokenSource tokenSource)
```
michelkaporin, line 616: What's the difference if you'd have the same method without ref argument, and pass token source without "ref" keyword?

Snyk.Code.Library/Service/SnykCodeFileProvider.cs
```diff
@@ -50,7 +50,10 @@ public string GetSolutionPath()
         public void RemoveFile(string file) => this.removedFiles.Add(file);
 
         /// <inheritdoc/>
-        public IEnumerable<string> GetChangedFiles() => this.changedFiles.Except(this.removedFiles);
+        public IEnumerable<string> GetChangedFiles() => this.changedFiles
+            .Except(this.removedFiles)
+            .Where(file => file != null)
+            .ToList();
```
michelkaporin, line 56: Is there a need for tolist here and in the method below?

michelkaporin (approved):

PR #63, opened 2021-12-04
Fix 🐛 VS2015 and VS2017 support [ROAD-372]

aldanchenko:
Added listener to load missing assemblies.


bastiandoetsch (approved):
Looks good afaict

PR #67, opened 2021-12-14
Fix: sync run buttons [ROAD-507]

aldanchenko:
Fixed disable/enable button on tool window main panel with run scan button on toolbar.


bastiandoetsch (approved):
LGTM

PR #68, opened 2021-12-14
Fix 🐛 broken settings issue [ROAD-466]

aldanchenko:
Fix issue with broken settings for project/solution. Now settings for solution saved and don't lost.

### DIFF
--- a/Snyk.VisualStudio.Extension/Snyk/VisualStudio/Extension/Settings/SnykSettingsLoader.cs
+++ b/Snyk.VisualStudio.Extension/Snyk/VisualStudio/Extension/Settings/SnykSettingsLoader.cs
@@ -1,0 +1,54 @@
+﻿namespace Snyk.VisualStudio.Extension.Settings
+{
+    using System;
+    using System.IO;
+    using System.Text;
+    using Snyk.Common;
+
+    /// <summary>
+    /// Load and save Snyk settings.
+    /// </summary>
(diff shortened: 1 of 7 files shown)


Snyk.VisualStudio.Extension/Snyk/VisualStudio/Extension/Settings/SnykSettingsLoader.cs
```diff
@@ -0,0 +1,54 @@
...
+
+        /// <summary>
+        /// Load <see cref="SnykSettings"/> instance.
+        /// </summary>
+        /// <returns>SnykSettings object.</returns>
+        public SnykSettings Load()
+        {
+            try
+            {
+                return Json.Deserialize<SnykSettings>(File.ReadAllText(this.GetSettingsFilePath(), Encoding.UTF8));
```
bastiandoetsch, line 26: Doesn't VS provide a settings storage itself? In IntelliJ you don't have to write/read a custom file.

bastiandoetsch (approved):

Snyk.VisualStudio.Extension/Snyk/VisualStudio/Extension/Settings/SnykSettingsLoader.cs
aldanchenko: Yes, it does. But very little functionality

PR #73, opened 2022-01-05
Fix: 🐛 scan large projects

aldanchenko:
Fix issues when scan large projects.

### DIFF
--- a/Snyk.Code.Library/Api/SnykCodeClient.cs
+++ b/Snyk.Code.Library/Api/SnykCodeClient.cs
@@ -19,8 +19,6 @@
         /// </summary>
         public const int MaxBundleSize = 4000000;
 
-        private const string LoginApiUrl = "login";
-
         private const string FiltersApiUrl = "filters";
 
         private const string BundleApiUrl = "bundle";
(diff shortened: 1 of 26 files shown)


Snyk.Code.Library/Api/SnykCodeClient.cs
```diff
@@ -21,8 +21,6 @@ public class SnykCodeClient : ISnykCodeClient
 
         private const string LoginApiUrl = "login";
```
michelkaporin, line 22: Is Login used anywhere?

Snyk.Code.Library/Service/AnalysisService.cs
```diff
@@ -62,7 +62,21 @@ public class AnalysisService : IAnalysisService
...
-                var analysisResultDto = await this.codeClient.GetAnalysisAsync(bundleHash, cancellationToken);
+                AnalysisResultDto analysisResultDto;
+
+                try
+                {
+                    analysisResultDto = await this.codeClient.GetAnalysisAsync(bundleHash, cancellationToken);
+                }
+                catch (Exception e)
+                {
+                    Logger.Error(e, "Error on try to get analysis.");
```
michelkaporin, line 73: What is the case that gets handled here? Does it makes sense to retry as waiting when it's our API client code that throws an exception?

Snyk.Code.Library/Service/BundleService.cs
```diff
@@ -61,11 +61,22 @@ public async Task<Bundle> CheckBundleAsync(string bundleId, CancellationToken ca
...
                 var pathToHashAndContentDict = codeCacheService.CreateFilePathToHashAndContentDictionary(resultBundle.MissingFiles);
 
-                await this.UploadFilesAsync(resultBundle.Id, pathToHashAndContentDict, fireScanCodeProgressUpdate, cancellationToken: cancellationToken);
+                try
+                {
+                    await this.UploadFilesAsync(resultBundle.Id, pathToHashAndContentDict, fireScanCodeProgressUpdate, cancellationToken: cancellationToken);
+                }
+                catch (Exception e)
+                {
+                    Logger.Error(e, "Error on upload missing files");
```
michelkaporin, line 70: We swallow an exception here. Does this mean we rely on retry for exception to go away? 
Can we have more documentation here to understand the use case?

Snyk.Code.Library/Service/BundleService.cs
```diff
@@ -61,11 +61,22 @@ public async Task<Bundle> CheckBundleAsync(string bundleId, CancellationToken ca
...
+                }
 
                 resultBundle = await this.CheckBundleAsync(bundle.Id, cancellationToken);
 
-                if (resultBundle.MissingFiles.Count() == 0)
+                var missingFiles = resultBundle.MissingFiles;
+
+                Logger.Warning("Missing files: {0}", string.Join(", ", missingFiles));
```
michelkaporin, line 77: Why is it warning severity and not info level? Isn't it normal to have missing file when checking for bundle?

michelkaporin:
You've switched to Newtonsoft.Json but Snyk.Code.Library is still using System.Text.Json. Can we make a switch there too to be consistent and reduce our maintenance efforts in perspective?

michelkaporin:
Quick iteration, thank you and great job @aldanchenko 👍

PR #76, opened 2022-01-13
fix: check before load settings file

aldanchenko:
Add check if settings file exists before load it.

### DIFF
--- a/Snyk.VisualStudio.Extension.Shared/Snyk/VisualStudio/Extension/Settings/SnykSettingsLoader.cs
+++ b/Snyk.VisualStudio.Extension.Shared/Snyk/VisualStudio/Extension/Settings/SnykSettingsLoader.cs
@@ -21,7 +21,14 @@
         {
             try
             {
-                return Json.Deserialize<SnykSettings>(File.ReadAllText(this.GetSettingsFilePath(), Encoding.UTF8));
+                string filePath = this.GetSettingsFilePath();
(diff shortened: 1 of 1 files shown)


Snyk.VisualStudio.Extension.Shared/Snyk/VisualStudio/Extension/Settings/SnykSettingsLoader.cs
```diff
@@ -21,7 +21,14 @@ public SnykSettings Load()
         {
             try
             {
-                return Json.Deserialize<SnykSettings>(File.ReadAllText(this.GetSettingsFilePath(), Encoding.UTF8));
+                string filePath = this.GetSettingsFilePath();
+
+                if (!File.Exists(filePath))
+                {
+                    return null;
```
michelkaporin, line 28: What happens if null is returned?

michelkaporin: So, nothing will break there? Why not to create an object here?

aldanchenko: No. It works without this changes. But if file not exists it will log error message. So added check just to clear log file

PR #78, opened 2022-01-13
fix: issues with tree view and snyk token

aldanchenko:
Fix snyk token for CLI.
Fix tree view in dark mode.


michelkaporin (approved):
@aldanchenko let's have more informative PR description. I know we had Slack message with little more info but it will be difficult to track it in future when looking at git commit history.

PR #82, opened 2022-01-25
[ROAD-351] Fix:  📺   screens workflow

aldanchenko:
This PR fix next issues:
* if snyk code disabled and os scan fail with error
* if no cli file show overview
* if cli file broken show error message in overview panel
* show overview after cli download
* download cli in settings and show run scan message

Screenshot 2022-01-21 at 20 09 19

Screenshot 2022-01-22 at 20 45 39

### DIFF
--- a/Snyk.VisualStudio.Extension.Shared/Settings/SnykGeneralSettingsUserControl.cs
+++ b/Snyk.VisualStudio.Extension.Shared/Settings/SnykGeneralSettingsUserControl.cs
@@ -67,46 +67,67 @@
             {
                 Logger.Information("Enter authenticate successCallback");
 
-                this.authProgressBar.Invoke((MethodInvoker)delegate
+                if (this.authProgressBar.IsHandleCreated)
                 {
-                    this.authProgressBar.Visible = false;
-                });
+                    this.authProgressBar.Invoke(new Action(() =>
+                    {
+                        this.authProgressBar.Visible = false;
+                    }));
+                }
 
-                this.tokenTextBox.Invoke((MethodInvoker)delegate
+                if (this.tokenTextBox.IsHandleCreated)
                 {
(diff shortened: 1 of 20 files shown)


Snyk.VisualStudio.Extension.Shared/Settings/SnykGeneralSettingsUserControl.cs
```diff
@@ -161,7 +181,7 @@ private void InitializeApiToken()
             {
                 string apiToken = this.NewCli().GetApiToken();
 
-                if (this.IsValidGuid(apiToken))
+                if (!string.IsNullOrEmpty(apiToken) && Common.Guid.IsValid(apiToken))
```
michelkaporin, line 184: perhaps we can move `string.IsNullOrEmpty(apiToken)` check to Guid.IsValid method?

Snyk.VisualStudio.Extension.Shared/Settings/SnykGeneralSettingsUserControl.cs
```diff
@@ -303,7 +319,7 @@ private void TokenTextBox_Validating(object sender, System.ComponentModel.Cancel
                 return;
             }
 
-            if (string.IsNullOrWhiteSpace(this.tokenTextBox.Text) || !this.IsValidGuid(this.tokenTextBox.Text))
+            if (string.IsNullOrWhiteSpace(this.tokenTextBox.Text) || !Common.Guid.IsValid(this.tokenTextBox.Text))
```
michelkaporin, line 322: do you need to check for `string.IsNullOrWhiteSpace(this.tokenTextBox.Text)` if Guid.IsValid does the same? IsNullOrWhiteSpace can be check in that method as well, if wanted.

michelkaporin (approved):

PR #83, opened 2022-01-26
[ROAD-322] Feat/add oss cache

aldanchenko:
Added cache for Open Source Security. If no changes in manifest files extension will return cached value. 

Screenshot 2022-01-26 at 15 22 45

### DIFF
--- a/Snyk.VisualStudio.Extension.Shared/Service/OssService.cs
+++ b/Snyk.VisualStudio.Extension.Shared/Service/OssService.cs
@@ -1,0 +1,68 @@
+﻿namespace Snyk.VisualStudio.Extension.Shared.Service
+{
+    using System.Threading;
+    using Serilog;
+    using Snyk.Common;
+    using Snyk.VisualStudio.Extension.Shared.CLI;
+    using Snyk.VisualStudio.Extension.Shared.Settings;
+
+    /// <inheritdoc/>
+    public class OssService : IOssService
+    {
+        private static readonly ILogger Logger = LogManager.ForContext<OssService>();
+
+        private SnykCli cli;
+
+        /// <inheritdoc/>
+        public bool IsCurrentScanProcessCanceled() => this.cli.ConsoleRunner.IsStopped;
+
+        /// <inheritdoc/>
+        public CliResult Scan(string path, ISnykOptions options, CancellationToken token)
+        {
+            token.ThrowIfCancellationRequested();
+
+            this.cli = new SnykCli
+            {
+                Options = options,
+            };
+
+            Logger.Information("Snyk Extension options");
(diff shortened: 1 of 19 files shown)


Snyk.VisualStudio.Extension.Shared/Service/OssService.cs
```diff
@@ -0,0 +1,86 @@
...
+            Logger.Information("Ignore Unknown CA is {IgnoreUnknownCA}", options.IgnoreUnknownCA);
+            Logger.Information("Additional Options is {AdditionalOptions}", options.AdditionalOptions);
+            Logger.Information("Is Scan All Projects is {IsScanAllProjects}", options.IsScanAllProjects);
+            Logger.Information("Solution path is {SolutionPath}", path);
+
+            Logger.Information("Start scan");
+
+            token.ThrowIfCancellationRequested();
+
+            var cliResult = this.cli.Scan(path);
```
bastiandoetsch, line 56: Aren't there any exceptions to be handled? (or to be sent to Sentry?)

Snyk.VisualStudio.Extension.Shared/Service/SnykTasksService.cs
```diff
@@ -363,72 +358,34 @@ private void ScanOss(FeaturesSettings featuresSettings)
...
-
                             Logger.Information("Scan finished");
                         }
-                        catch (Exception e)
+                        catch (OssScanException e)
                         {
-                            bool isCancelled = this.cli.ConsoleRunner.IsStopped;
-
-                            this.cli = null;
+                            Logger.Error(e, "Oss scan exception");
```
bastiandoetsch, line 381: Is the logger is connected to Sentry via a Serilog-SInk? Just asking...

bastiandoetsch (approved):

Snyk.VisualStudio.Extension.Shared/Service/OssService.cs
aldanchenko: Yes, this try/catch/finally exists on upper level - https://github.com/snyk/snyk-visual-studio-plugin/blob/102ca8fd97194e512c44780f90eb19254527d9e8/Snyk.VisualStudio.Extension.Shared/Service/SnykTasksService.cs#L371

Snyk.VisualStudio.Extension.Shared/Service/SnykTasksService.cs
aldanchenko: Yes, Logger and Sentry configuration in one place here - https://github.com/snyk/snyk-visual-studio-plugin/blob/102ca8fd97194e512c44780f90eb19254527d9e8/Snyk.Common/LogManager.cs#L45

And LogManager used across all extension.

PR #84, opened 2022-01-27
[ROAD-658] feat: add user, project type, vs version and extension version to Sen…

aldanchenko:
Added user, project type, vs version and extension version to Sentry events.

### DIFF
--- a/Snyk.Common/LogManager.cs
+++ b/Snyk.Common/LogManager.cs
@@ -50,6 +50,13 @@
             sentryConfiguration.BeforeSend = beforeSendHookFunc;
         }
 
+        /// <summary>
+        /// Check is sentry configuration already initialized.
+        /// If sentryConfiguration not null it means logger configuration already created and ready to use.
+        /// </summary>
+        /// <returns>True if logger configuration already created</returns>
+        public static bool IsInitialized() => sentryConfiguration != null;
+
         private static Logger CreateLogger() => new LoggerConfiguration()
                 .Enrich.WithProcessId()
                 .Enrich.WithThreadId()
--- a/Snyk.Common/LogManager.cs
+++ b/Snyk.Common/LogManager.cs
@@ -2,6 +2,8 @@
 {
     using System;
     using System.IO;
+    using Sentry;
+    using Sentry.Serilog;
     using Serilog;
     using Serilog.Core;
     using Serilog.Events;
@@ -28,6 +30,8 @@
 
         private static LoggingLevelSwitch loggingLevelSwitch = new LoggingLevelSwitch(defaultLoggingLevel);
 
+        private static SentrySerilogOptions sentryConfiguration;
+
         private static Lazy<Logger> Logger { get; } = new Lazy<Logger>(CreateLogger);
 
         /// <summary>
@@ -36,6 +40,15 @@
         /// <typeparam name="T">Type of class.</typeparam>
         /// <returns><see cref="ILogger"/> implementation for class.</returns>
         public static ILogger ForContext<T>() => ForContext(typeof(T));
+
+        /// <summary>
+        /// Set Sentry BeforeSend hook function.
+        /// </summary>
+        /// <param name="beforeSendHookFunc">Function for execute before Sentry send events.</param>
+        public static void SetSentryBeforeSendHook(Func<SentryEvent, SentryEvent> beforeSendHookFunc)
+        {
+            sentryConfiguration.BeforeSend = beforeSendHookFunc;
+        }
 
         private static Logger CreateLogger() => new LoggerConfiguration()
                 .Enrich.WithProcessId()
@@ -46,10 +59,14 @@
                 {
                     var appSettings = SnykExtension.GetAppSettings();
 
+                    config.Release = SnykExtension.GetIntegrationVersion();
+
                     config.Environment = appSettings.Environment;
                     config.Dsn = appSettings.SentryDsn;
 
                     config.AttachStacktrace = true;
+
+                    sentryConfiguration = config;
                 })
                 .WriteTo.File(
                     Path.Combine(SnykDirectory.GetSnykAppDataDirectoryPath(), "snyk-extension.log"),
--- a/Snyk.Common/LogManager.cs
+++ b/Snyk.Common/LogManager.cs
@@ -2,7 +2,6 @@
 {
     using System;
     using System.IO;
-    using Sentry;
     using Sentry.Serilog;
     using Serilog;
     using Serilog.Core;
@@ -30,7 +29,7 @@
 
         private static LoggingLevelSwitch loggingLevelSwitch = new LoggingLevelSwitch(defaultLoggingLevel);
 
-        private static SentrySerilogOptions sentryConfiguration;
+        public static SentrySerilogOptions SentryConfiguration { get; private set; }
 
         private static Lazy<Logger> Logger { get; } = new Lazy<Logger>(CreateLogger);
 
@@ -40,22 +39,6 @@
         /// <typeparam name="T">Type of class.</typeparam>
         /// <returns><see cref="ILogger"/> implementation for class.</returns>
         public static ILogger ForContext<T>() => ForContext(typeof(T));
-
-        /// <summary>
-        /// Set Sentry BeforeSend hook function.
-        /// </summary>
-        /// <param name="beforeSendHookFunc">Function for execute before Sentry send events.</param>
-        public static void SetSentryBeforeSendHook(Func<SentryEvent, SentryEvent> beforeSendHookFunc)
-        {
-            sentryConfiguration.BeforeSend = beforeSendHookFunc;
-        }
-
-        /// <summary>
-        /// Check is sentry configuration already initialized.
-        /// If sentryConfiguration not null it means logger configuration already created and ready to use.
-        /// </summary>
-        /// <returns>True if logger configuration already created</returns>
-        public static bool IsInitialized() => sentryConfiguration != null;
 
         private static Logger CreateLogger() => new LoggerConfiguration()
                 .Enrich.WithProcessId()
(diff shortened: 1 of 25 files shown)


Snyk.Common/SnykExtension.cs
```diff
@@ -72,9 +72,12 @@ public static string GetIntegrationVersion()
                 }
                 catch (Exception e)
                 {
-                    var logger = LogManager.ForContext<SnykExtension>();
+                    if (LogManager.IsInitialized())
```
bastiandoetsch, line 75: Do we want to ignore the Exception if the LogManager is not initialized? Shouldn't we at least report it to the user or write it to standard error or such?

Snyk.VisualStudio.Extension.2022/Snyk.VisualStudio.Extension.2022.csproj
```diff
@@ -84,6 +84,9 @@
     <PackageReference Include="Markdig">
       <Version>0.24.0</Version>
     </PackageReference>
+    <PackageReference Include="Sentry">
+      <Version>3.12.1</Version>
```
bastiandoetsch, line 88: Latest version seems to be 3.13.0 - consider updating :). (https://docs.sentry.io/platforms/dotnet/)

Snyk.VisualStudio.Extension.Shared/Service/SnykService.cs
```diff
@@ -247,6 +246,21 @@ public string GetApiToken()
...
+        {
+            var shell = await this.Package.GetServiceAsync(typeof(SVsShell)) as IVsShell;
+            shell.GetProperty((int)__VSSPROPID5.VSSPROPID_ReleaseVersion, out object vsVersion);
+
+            SentrySdk.ConfigureScope(scope =>
+            {
+                scope.User = new User { Id = this.AnalyticsService.UserIdAsHash, };
+
+                scope.SetTag("vs.version", vsVersion.ToString());
+                scope.SetTag("vs.edition", dte.Edition);
```
bastiandoetsch, line 260: Do we also have the plugin version available somewhere and can put it here? Or is that already done elsewhere?

Snyk.VisualStudio.Extension.Shared/SnykAnalytics/SnykAnalyticsService.cs
```diff
@@ -78,6 +80,22 @@ public static SnykAnalyticsService Instance
...
+        /// <summary>
+        /// Gets a value indicating whether user id hash string.
+        /// </summary>
+        public string UserIdAsHash
+        {
+            get
+            {
+                if (string.IsNullOrEmpty(this.userIdAsHash) && !string.IsNullOrEmpty(this.UserId))
+                {
+                    this.userIdAsHash = Sha256.ComputeHash(this.UserId);
```
bastiandoetsch, line 92: I love this 💜. Hashing the userId feels so good :).

Snyk.VisualStudio.Extension.Shared/SnykAnalytics/SnykAnalyticsService.cs
```diff
@@ -207,7 +226,7 @@ private void Identify()
             this.analyticsClient?.Identify(this.anonymousUserId, new Traits());
```
bastiandoetsch, line 226: What happens, if the analytics client throws an exception? My adblocking e.g. sometimes blocks websites like Sentry or Amplitude.

Snyk.Common/SnykExtension.cs
aldanchenko: Not sure if we should inform a user that the logger is not working for us. I think that if everything else works, then it is more important for a user. And this exception must appear only for test case. If user run extension in VS it always have manifest file and version number.

Snyk.VisualStudio.Extension.Shared/Service/SnykService.cs
aldanchenko: Yes, we do it in this place - https://github.com/snyk/snyk-visual-studio-plugin/blob/33fe6d824951045e28bbc6e66b8babdcace1796b/Snyk.Common/LogManager.cs#L69

Snyk.VisualStudio.Extension.Shared/Service/SnykService.cs
```diff
@@ -247,6 +246,21 @@ public string GetApiToken()
             return this.NewCli().GetApiToken();
         }
 
+        /// <inheritdoc/>
+        public async Task SetupSentryAsync()
+        {
+            var shell = await this.Package.GetServiceAsync(typeof(SVsShell)) as IVsShell;
+            shell.GetProperty((int)__VSSPROPID5.VSSPROPID_ReleaseVersion, out object vsVersion);
+
+            SentrySdk.ConfigureScope(scope =>
```
michelkaporin, line 255: This violates single responsibility principle. Consider moving configuration in Sentry-related class.

Snyk.VisualStudio.Extension.Shared/SnykAnalytics/SnykAnalyticsService.cs
aldanchenko: As I understand there won't be any errors, extension just keep working

Snyk.VisualStudio.Extension.Shared/Service/SnykSolutionService.cs
```diff
@@ -190,6 +190,33 @@ public string GetPath()
             return solutionPath;
         }
 
+        /// <inheritdoc/>
+        public string GetProjectType()
```
michelkaporin, line 194: Where is this used?

Snyk.VisualStudio.Extension.Shared/Service/SnykVsSolutionLoadEvents.cs
```diff
@@ -213,5 +229,15 @@ public int OnQueryBackgroundLoadProjectBatch(out bool pfShouldDelayLoadToNextIdl
         /// <param name="pfCancel">Cancel.</param>
         /// <returns>VSConstants.S_OK.</returns>
         public int OnQueryUnloadProject(IVsHierarchy pRealHierarchy, ref int pfCancel) => VSConstants.S_OK;
+
+        private void SetupSentrySolutionInformation(SolutionType solutionType)
+        {
+            LogManager.SetSentryBeforeSendHook(sentryEvent =>
+            {
+                sentryEvent.SetTag("vs.project.type", solutionType.ToString());
```
michelkaporin, line 237: I'd move this configuration into LogManager instead or Sentry-related class better. Having tags scattered across different classes provide less visibility on how Sentry events are enriched. Consider creating a separate class for Sentry that will provide all the necessary info to LogManager and expose required method to set required Sentry context configuration.

I'd expect here to have a call to it, smth like:
`Sentry.SetProjectType(solutionType)`

Snyk.VisualStudio.Extension.Shared/SnykAnalytics/SnykAnalyticsService.cs
```diff
@@ -78,6 +80,22 @@ public static SnykAnalyticsService Instance
...
+        public string UserIdAsHash
+        {
+            get
+            {
+                if (string.IsNullOrEmpty(this.userIdAsHash) && !string.IsNullOrEmpty(this.UserId))
+                {
+                    this.userIdAsHash = Sha256.ComputeHash(this.UserId);
+                }
+
+                return this.userIdAsHash;
```
michelkaporin, line 95: Why do we default to the empty or null Id here, if UserId is not set? Perhaps returning anonymous would be more sensible?

Also it would be great to persist anonymousId across different VS sessions (as we do in other plugins), since you're generating it new every time.

Snyk.Common/LogManager.cs
```diff
@@ -37,6 +41,22 @@ public static class LogManager
...
+        {
+            sentryConfiguration.BeforeSend = beforeSendHookFunc;
+        }
+
+        /// <summary>
+        /// Check is sentry configuration already initialized.
+        /// If sentryConfiguration not null it means logger configuration already created and ready to use.
+        /// </summary>
+        /// <returns>True if logger configuration already created</returns>
+        public static bool IsInitialized() => sentryConfiguration != null;
```
michelkaporin, line 58: This definitely doesn't belong to this class. Consider having encapsulating error reporting-related logic in a separate class 👍

Snyk.VisualStudio.Extension.Shared/Service/ISnykServiceProvider.cs
```diff
@@ -99,5 +99,11 @@ public interface ISnykServiceProvider
         /// <param name="serviceType">Service type.</param>
         /// <returns>VS service instance.</returns>
         Task<object> GetServiceAsync(Type serviceType);
+
+        /// <summary>
+        /// Setup Sentry scope with VS version, edition and user.
+        /// </summary>
+        /// <returns>Task.</returns>
+        System.Threading.Tasks.Task SetupSentryAsync();
```
michelkaporin, line 107: Namespace prefix could be moved to using, to make it more readable

michelkaporin (approved):

aldanchenko:
@michelkaporin this PR ready for review. Last changes include your suggestions. Thanks!

Snyk.VisualStudio.Extension.Shared/SnykAnalytics/SnykAnalyticsService.cs
```diff
@@ -88,37 +102,45 @@ public static SnykAnalyticsService Instance
...
+        /// <param name="options">Snyk options.</param>
+        public static void Initialize(ISnykOptions options)
         {
             try
             {
                 SnykAppSettings appSettings = SnykExtension.GetAppSettings();
 
+                if (string.IsNullOrEmpty(options.SentryAnonymousUserId))
+                {
+                    options.SentryAnonymousUserId = System.Guid.NewGuid().ToString();
```
michelkaporin, line 116: Is anonymous id persisted between different sessions, if user restart their VS instance? If not, we should store it. If it's hard let's have a separate ticket to address that (could also be a part of Iteratively integration ticket).

michelkaporin (approved):

michelkaporin:
Minor things left but otherwise looks good ✅ Love the separation of concerns that we could achieve here with latest changes 💪 great job!

Snyk.VisualStudio.Extension.Shared/SnykAnalytics/SnykAnalyticsService.cs
aldanchenko: Yes, this part `options.SentryAnonymousUserId` will save it in settings. And next time it will try to get it from settings.

PR #85, opened 2022-01-31
[ROAD-682] fix: add check is sentry error related to extension

aldanchenko:
To fix issue with not related to extension exceptions now it filter Sentry errors before send. It check Logger property and StackTrace content. 

Errors from extension must have SentryEvent.Logger = "Snyk.VisualStudio.CLI.SomeClass" property set. And also error must have Snyk mention in StackTrace

### DIFF
--- a/Snyk.Common/Sentry/AbstractIntegration.cs
+++ b/Snyk.Common/Sentry/AbstractIntegration.cs
@@ -1,0 +1,46 @@
+﻿namespace Snyk.Common.Sentry
+{
+    using System;
+    using global::Sentry;
+    using global::Sentry.Integrations;
+
+    /// <summary>
+    /// Snyk integration abstract class for Sentry.
+    /// </summary>
+    public abstract class AbstractIntegration : ISdkIntegration
+    {
+        protected const string HandledKey = "Sentry:Handled";
+
+        protected const string MechanismKey = "Sentry:Mechanism";
+
+        protected const string SnykKey = "Snyk";
+
+        protected readonly IAppDomain appDomain;
+
+        protected IHub hub;
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="AbstractIntegration"/> class.
+        /// Initialize appDoamin property with <see cref="AppDomainAdapter"/>.
+        /// </summary>
+        public AbstractIntegration() => this.appDomain = AppDomainAdapter.Instance;
+
+        /// <summary>
+        /// Registers this integration with the hub.
+        /// </summary>
+        /// <remarks>
+        /// This method is invoked when the Hub is created.
+        /// </remarks>
+        /// <param name="hub">The hub.</param>
+        /// <param name="options">The options.</param>
+        public abstract void Register(IHub hub, SentryOptions options);
+
+        /// <summary>
+        /// Check is provided exception stack trace contains mentions about extension.
+        /// </summary>
+        /// <param name="e">Source exception.</param>
+        /// <returns>True if contains mentions about Snyk extension.</returns>
+        protected static bool IsNeedToHandleException(Exception e)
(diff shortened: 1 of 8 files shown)


michelkaporin (changes requested):
Let's work with Sentry integrations instead, as discussed during the call today 👌

Snyk.Common/Sentry/AbstractIntegration.cs
```diff
@@ -0,0 +1,46 @@
...
+        /// <param name="options">The options.</param>
+        public abstract void Register(IHub hub, SentryOptions options);
+
+        /// <summary>
+        /// Check is provided exception stack trace contains mentions about extension.
+        /// </summary>
+        /// <param name="e">Source exception.</param>
+        /// <returns>True if contains mentions about Snyk extension.</returns>
+        protected static bool IsNeedToHandleException(Exception e)
+            => e != null && e.StackTrace != null && e.StackTrace.Contains(SnykKey);
```
bastiandoetsch, line 44: Maybe it would be good to be case-insensitive here, so we catch everything Snyk, and not only the capitalized parts.

bastiandoetsch (approved):
Please consider to do case-insensitive handling. Apart from that, LGTM.

Snyk.Common/Sentry/AbstractIntegration.cs
michelkaporin, line 44: Does stacktrace include path to the extension? Perhaps we can be more precise and say the path has to be within the stacktrace, instead just "Snyk" string?

Snyk.Common/Sentry/AbstractIntegration.cs
```diff
@@ -0,0 +1,46 @@
...
+        /// <param name="hub">The hub.</param>
+        /// <param name="options">The options.</param>
+        public abstract void Register(IHub hub, SentryOptions options);
+
+        /// <summary>
+        /// Check is provided exception stack trace contains mentions about extension.
+        /// </summary>
+        /// <param name="e">Source exception.</param>
+        /// <returns>True if contains mentions about Snyk extension.</returns>
+        protected static bool IsNeedToHandleException(Exception e)
```
michelkaporin, line 43: Does this work when AggregateException with N-layers of InnerException comes?

Snyk.Common/Sentry/AbstractIntegration.cs
aldanchenko: Discussed in Slack. Decided to leave as is right now.

Snyk.Common/Sentry/AbstractIntegration.cs
aldanchenko: Good point @michelkaporin!

aldanchenko: @michelkaporin changed this part according to your comment. Please, have a look ;)

michelkaporin: Looks good. Consider using Linq to simplify this simple foreach :)

aldanchenko:
```
protected static bool IsNeedToHandleException(Exception e)
        {
            if (IsExceptionRelatedToExtension(e))
            {
                return true;
            }

            if (e is AggregateException aggregateException 
                && aggregateException.Flatten().InnerExceptions.Where(ex => IsExceptionRelatedToExtension(ex)).IsNullOrEmpty())
            {
                return true;
            }

            return false;
        }
```

aldanchenko: @michelkaporin this will be ok?

michelkaporin: You can work with .Any() to have it simpler 👍

michelkaporin (approved):

PR #89, opened 2022-02-07
[ROAD-721] fix: ide crush if select oss critical vulnerability issue

aldanchenko:
VS2017 crush on OSS vulnerability select (with critical severity). Unfortunately mkdig library crash on parse markdown in VS2017. Replaced mkdig with MarkdownSharp library.


bastiandoetsch (approved):
LGTM

PR #100, opened 2022-03-08
fix: Visual Studio 2022 supports amd64 only

michelkaporin:
When publishing 2022, it fails with the following message:
```
Error: 
(Snyk Vulnerability Scanner) Target 'Microsoft.VisualStudio.Community' with version range '[17.0,18.0)' specifies 'x86' as the target product architecture, but Visual Studio 2022 and above only supports `amd64`. More info at http://aka.ms/VS2022ExtensionSupport 
```

This PR addresses the problem.


aldanchenko (commented):
LGTM 👍

PR #102, opened 2022-03-14
[ROAD-641] Fix: deployment pipeline for vs

aldanchenko:
Fix calculate vsix version using tag name. Now it will get tag name and increase path number and use it as new vsix version.


pavel-snyk (approved):
LGTM, nice trick with injecting version from the action itself.

.github/workflows/release.yml
```diff
@@ -55,54 +54,80 @@ jobs:
...
-    - name: Increment VSIX version
+    - name: Calculate next Git tag
       id: vsix_version
-      uses: timheuer/vsix-version-stamp@v1
+      uses: ./.github/actions/next-git-tag -Verbose
+
+    - name: Set VSIX version for 2015-2019
+      uses: cezarypiatek/VsixVersionAction@1.0
       with:
-        manifest-file: ${{ env.VsixManifestPath }}
```
aldanchenko, line 62: It seems this VsixManifestPath not used in release.yml. Maybe we could remove it?

.github/workflows/release.yml
```diff
@@ -55,54 +54,80 @@ jobs:
...
-    
+
+    - name: Upload Release 2022 Asset
+      uses: actions/upload-release-asset@v1
+      env:
+        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      with:
+        upload_url: ${{ steps.create_release.outputs.upload_url }}
+        asset_path: .\Snyk.VisualStudio.Extension.2022\bin\Release\Snyk.VisualStudio.Extension.vsix
+        asset_name: Snyk_Vulnerability_Scanner-${{ steps.vsix_version.outputs.next-tag }}.vsix
```
aldanchenko, line 118: Maybe we need to add here 2022 or something in vsix name to separate 2015-2019 and 2022 vsix file?

.github/workflows/release.yml
pavel-snyk: @aldanchenko , it's removed if you take a look on the diff :)
image

.github/workflows/release.yml
pavel-snyk: good catch, otherwise it will be overriden

pavel-snyk: done with 39ebb66fabaf442bf664efdb5b2ee4fe8aff034c

.github/workflows/release.yml
```diff
@@ -55,54 +54,80 @@ jobs:
     - name: Restore NuGet packages
       run: nuget restore ${{env.SOLUTION_FILE_PATH}}
 
-    - name: Increment VSIX version
+    - name: Calculate next Git tag
```
aldanchenko, line 57: [nitpick] maybe add some more description in name. Like
"Calculate next Git tag (vsix version)". But I don't sure

.github/workflows/release.yml
```diff
@@ -55,54 +54,80 @@ jobs:
...
     - name: Build
       run: |
         msbuild ${{env.SOLUTION_FILE_PATH}} /p:configuration=Release /p:DeployExtension=false /p:ZipPackageCompressionLevel=normal /v:m
       shell: powershell
 
     - name: Tests
       run: vstest.console.exe **\*.Tests.dll
-      
+
+    - name: Set up Git actions user
```
aldanchenko, line 81: [question] why we need this?

.github/workflows/release.yml
```diff
@@ -55,54 +54,80 @@ jobs:
...
+      uses: fregante/setup-git-user@v1
+
+    - name: Push Git tag
+      run: |
+        git tag ${{ steps.vsix_version.outputs.next-tag }}
+        git push origin main
+        git push origin main --tags
+
     - name: Create Release
```
aldanchenko, line 90: [nitpick] Maybe change name to "Create Git tag release" or something similar because we have few different releases (vsix release, tags release)

.github/workflows/release.yml
aldanchenko: Thanks for added 2022. 
[nitpick] I recheck how we name 2022 vsix version. Previously I used `"Snyk_Vulnerability_Scanner-1.1.1.2022.vsix"` Maybe for consistency we could use this format

aldanchenko:
@pavel-snyk super nice! 👍

.github/workflows/release.yml
pavel-snyk: Git user name and email are empty without it, so `git tag ...` command in the next step would fail.

.github/workflows/release.yml
pavel-snyk: done with 7aa1f3ef2a59f2c9c9006714c8cbf05d623c920d

.github/workflows/release.yml
pavel-snyk: done with 774f4880d20a7835196c44f2d8a8a6e448226de4

pavel-snyk (approved):
😉 LGTM.

PR #105, opened 2022-03-20
[ROAD-774] 🐛 Fix: project settings not persisted

aldanchenko:
This PR fix issue with save project settings (settings related to current open solution).

Current implementation will use solution path as key for additional settings map.


michelkaporin:
@aldanchenko build seem to fail.
- Can you fix before the review?
- Also can you explain why settings were not persisted previously and are persisted with your solution now?

aldanchenko:
About this
>Also can you explain why settings were not persisted previously and are persisted with your solution now?

For save project settings SnykUserStorageSettingsService.cs use a map with project name -> project options pair. But not for all projects [this method](https://github.com/snyk/snyk-visual-studio-plugin/blob/28de363d2dd371b125821cebc87f9774e6704c3d/Snyk.VisualStudio.Extension.Shared/Settings/SnykUserStorageSettingsService.cs#L337) return unique project name. For some projects it return "misc" string or null. 

Current solution: use project path instead project name (from vs api).

aldanchenko:
@michelkaporin tests fixed. Ready for review

README.md
```diff
@@ -156,7 +156,7 @@ After the plugin is installed, you can set the following configurations for the
 - **Ignore unknown CA**: Ignore unknown certificate authorities.
 - **Organization**: Specify the ORG_NAME to run Snyk commands tied to a specific organization.
 - **Send usage analytics**: To help us improve the extension, you can let your Visual Studio send us information about how it’s working.
-- **Project settings**: Specify any additional Snyk CLI parameters.
+- **Solution settings**: Specify any additional Snyk CLI parameters.
```
michelkaporin, line 159: Nice spot!

Snyk.VisualStudio.Extension.Shared/Settings/SnykUserStorageSettingsService.cs
```diff
@@ -363,5 +308,7 @@ private SnykSettings LoadSettings()
 
             return settings;
         }
+
+        private int GetSolutionPathHash() => this.solutionService.GetPath().FirstCharToLowerCase().GetHashCode();
```
michelkaporin, line 312: Since Windows has case insensitive path, can we lowercase the **whole project path**, not the first char instead?

michelkaporin (approved):
LGTM, but let's address the last comment.

Snyk.VisualStudio.Extension.Shared/Settings/SnykUserStorageSettingsService.cs
aldanchenko: Good point!

Snyk.Common/StringExtensions.cs
```diff
@@ -14,22 +14,21 @@ public static class StringExtensions
...
         /// <param name="source">Source string.</param>
         /// <returns>Result string.</returns>
         public static string RemoveTrailingSlashes(this string source) => Regex.Replace(source, "/+$", string.Empty);
+
+        /// <summary>
+        /// To lower case first char of string.
+        /// </summary>
+        /// <param name="str">Source string.</param>
+        /// <returns>Result string with lower case first char.</returns>
+        public static string FirstCharToLowerCase(this string str)
```
michelkaporin, line 31: do you still need this? :)

michelkaporin (dismissed):

pavel-snyk (approved):
@aldanchenko, 👍 as discussed internally.

PR #109, opened 2022-03-28
[ROAD-659] Fix: sentry error parameter is incorrect

aldanchenko:
In this PR reworked logic for get files in Solution. Now extension get solution files using VS toolkit api. 

First extension will check project/solution type. If project opened as folder it will get folder files using .net classes. If project is .net solution it will use VS toolkit api.

If project not contains files to scan it will show message in corresponding tree node item. For example

Screenshot 2022-03-28 at 08 44 38

### DIFF
--- a/Snyk.Code.Library/Service/SnykCodeService.cs
+++ b/Snyk.Code.Library/Service/SnykCodeService.cs
@@ -2,6 +2,7 @@
 {
     using System;
     using System.Collections.Generic;
+    using System.Linq;
     using System.Threading;
     using System.Threading.Tasks;
     using Serilog;
@@ -115,7 +116,14 @@
         {
             this.FireScanProgressEvent(SnykCodeScanState.Preparing, 0);
 
-            var files = await this.GetFilteredFilesAsync(fileProvider.GetSolutionPath(), fileProvider.GetFiles());
+            var solutionFiles = await fileProvider.GetFilesAsync();
+
+            var files = await this.GetFilteredFilesAsync(fileProvider.GetSolutionPath(), solutionFiles);
+
+            if (files == null || files.Count() == 0)
+            {
+                return null;
+            }
 
             this.codeCacheService.Initialize(files);
 
--- a/Snyk.VisualStudio.Extension.Shared/Service/SnykSolutionService.cs
+++ b/Snyk.VisualStudio.Extension.Shared/Service/SnykSolutionService.cs
@@ -3,14 +3,17 @@
     using System;
     using System.Collections.Generic;
     using System.IO;
+    using System.Linq;
     using EnvDTE;
     using Microsoft.VisualStudio;
+    using Microsoft.VisualStudio.Shell;
     using Microsoft.VisualStudio.Shell.Interop;
     using Serilog;
     using Snyk.Code.Library.Service;
     using Snyk.Common;
     using Snyk.VisualStudio.Extension.Shared.Service;
     using Task = System.Threading.Tasks.Task;
+    using Toolkit = Community.VisualStudio.Toolkit;
 
     /// <summary>
     /// Incapsulate logic for work with Visual Studio solutions.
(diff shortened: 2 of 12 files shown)


Snyk.VisualStudio.Extension.Shared/Service/SnykSolutionService.cs
```diff
@@ -261,5 +232,71 @@ private IList<string> GetSolutionDirectoryFiles()
...
+
+        private async System.Threading.Tasks.Task<IList<string>> GetSolutionItemFilesAsync(IEnumerable<Toolkit.SolutionItem> solutionItems)
+        {
+            await ThreadHelper.JoinableTaskFactory.SwitchToMainThreadAsync();
+
+            var files = new List<string>();
+
+            foreach (var item in solutionItems)
+            {
+                var itemFiles = await this.GetSolutionItemFilesAsync(item);
```
michelkaporin, line 244: This will block the thread. We should fetch files asynchronously instead. Consider using Task.WhenAll to await afterwards.

Snyk.VisualStudio.Extension.Shared/Service/SnykSolutionService.cs
```diff
@@ -208,17 +178,18 @@ public string GetPath()
...
+            // If solution is folder type when just get all files in solution directory.
+            if (this.IsSolutionFolderType())
             {
-                solutionProjectsFiles = this.GetSolutionDirectoryFiles();
+                return this.GetSolutionDirectoryFiles();
             }
 
-            return solutionProjectsFiles;
+            // If Solution is projects type when get all files in solution projects.
+            return await this.GetSolutionFilesAsync();
```
michelkaporin, line 192: Not sure what is "project type" solution. Maybe let's sync on that offline :)

Snyk.VisualStudio.Extension.Shared/Service/SnykSolutionService.cs
```diff
@@ -261,5 +232,71 @@ private IList<string> GetSolutionDirectoryFiles()
...
+
+            foreach (var item in solutionItem.Children)
+            {
+                if (item.Type == Toolkit.SolutionItemType.PhysicalFile)
+                {
+                    files.Add(item.FullPath);
+                }
+                else
+                {
+                    var itemFiles = await this.GetSolutionItemFilesAsync(item);
```
michelkaporin, line 274: We should unify some code between this and the above method, if possible.

Snyk.Code.Library/Service/SnykCodeService.cs
```diff
@@ -115,7 +116,14 @@ private async Task<AnalysisResult> NewScanAsync(IFileProvider fileProvider, Canc
...
             this.FireScanProgressEvent(SnykCodeScanState.Preparing, 0);
 
-            var files = await this.GetFilteredFilesAsync(fileProvider.GetSolutionPath(), fileProvider.GetFiles());
+            var solutionFiles = await fileProvider.GetFilesAsync();
+
+            var files = await this.GetFilteredFilesAsync(fileProvider.GetSolutionPath(), solutionFiles);
+
+            if (files == null || files.Count() == 0)
+            {
+                throw new NoFilesForScanException();
```
michelkaporin, line 125: We shouldn't throw an exception, because exceptions represent errors that occur during application execution. This case is however a valid case in which application can end up.

Every exception that gets thrown has also additional performance impact, because it does generate a stack trace, when thrown. 

Instead, we should be communicating this back to the caller via return parameter.

Snyk.VisualStudio.Extension.Shared/Service/SnykSolutionService.cs
```diff
@@ -93,40 +95,8 @@ public async Task InitializeAsync(ISnykServiceProvider serviceProvider)
...
-            }
-            catch (Exception e)
-            {
-                Logger.Error(e, "Failed to get project files");
-            }
-
-            return solutionFiles;
-        }
+        /// <inheritdoc/>
+        public bool IsSolutionFolderType() => Directory.Exists(this.ServiceProvider.DTE.Solution.FullName);
```
michelkaporin, line 99: Let's document what `this.ServiceProvider.DTE.Solution.FullName` can be.

michelkaporin, line 99: `IsSolutionOpenedAsFolder()`

Snyk.VisualStudio.Extension.Shared/Service/SnykSolutionService.cs
```diff
@@ -208,17 +178,18 @@ public string GetPath()
...
-            if (solutionProjectsFiles.Count == 0)
+            // If solution is folder type when just get all files in solution directory.
+            if (this.IsSolutionFolderType())
             {
-                solutionProjectsFiles = this.GetSolutionDirectoryFiles();
+                return this.GetSolutionDirectoryFiles();
             }
 
-            return solutionProjectsFiles;
+            // If Solution is projects type when get all files in solution projects.
```
michelkaporin, line 191: If normal solution, then get all files in solution projects.

Snyk.VisualStudio.Extension.Shared/Service/SnykSolutionService.cs
aldanchenko: The above method was deleted.

Snyk.VisualStudio.Extension.Shared/Service/SnykSolutionService.cs
```diff
@@ -261,5 +235,56 @@ private IList<string> GetSolutionDirectoryFiles()
...
+
+        private async System.Threading.Tasks.Task<IList<string>> GetSolutionItemFilesAsync(Toolkit.SolutionItem solutionItem)
+        {
+            await ThreadHelper.JoinableTaskFactory.SwitchToMainThreadAsync();
+
+            try
+            {
+                var files = new List<string>();
+
+                foreach (var item in solutionItem.Children)
```
michelkaporin, line 247: Does this loop take long time for big projects? We can optimise await inside, if needed.

michelkaporin (approved):

Snyk.VisualStudio.Extension.Shared/Service/SnykSolutionService.cs
aldanchenko: Testing it. For now don't see any speed loose. But it's without any numbers. Just personal feelings

PR #110, opened 2022-03-29
[ROAD-818] fix: handle file content changes and invalidate cache

aldanchenko:
Fix invalidate cache for snykcode and oss if file content (text) changed


pavel-snyk (approved):
LGTM, discussed in Slack.

PR #111, opened 2022-03-30
fix: restore proper logic for solution open check

aldanchenko:
Restore proper logic for check is solution open or not.


pavel-snyk (approved):
LGTM, clarified internally.

PR #114, opened 2022-04-01
[ROAD-823] fix: invalidate cache if solution switch

aldanchenko:
Fix issue with cache invalidate if solution switch to another. For this added clear OSS cache method call on different solution events.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,4 +3,5 @@
 ## [1.1.7]
 
 ### Fixed
-- Issue that brought Visual Studio 2017 to crash
+- Issue that brought Visual Studio 2017 to crash.
+- Issuw with cache invalidation for Open Source vulnerabilities.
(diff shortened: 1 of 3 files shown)


michelkaporin (commented):
Can we also test this, since it's a bug fix ? :-)

aldanchenko:
@michelkaporin don't sure because this functionality related to vs sdk.

michelkaporin (approved):
@aldanchenko you could potentially mock the solution service and invoke event methods, then verifying if `Clean()` was called on the solution service.

Since it's probably too much for a unit test, we can leave this out, unless you find it's still worth testing. Approving for now. 

Don't forget to update changelog with this fix.

CHANGELOG.md
```diff
@@ -3,4 +3,5 @@
 ## [1.1.7]
 
 ### Fixed
-- Issue that brought Visual Studio 2017 to crash
\ No newline at end of file
+- Issue that brought Visual Studio 2017 to crash.
+- Issuw with cache invalidation for Open Source vulnerabilities.
```
michelkaporin, line 7: Issue

michelkaporin (approved):

PR #115, opened 2022-04-03
[ROAD-816] fix: multiple windows opened at once

asaf92:
Fixed an issue where clicking the CVSS link would open more than one link

### DIFF
--- a/Snyk.VisualStudio.Extension.Shared/UI/Toolwindow/DescriptionHeaderPanel.xaml.cs
+++ b/Snyk.VisualStudio.Extension.Shared/UI/Toolwindow/DescriptionHeaderPanel.xaml.cs
@@ -79,15 +79,14 @@
 
                 this.cvssLink.NavigateUri = new Uri(vulnerability.Url);
                 this.cvssLinkText.Text = vulnerability.Id;
-                this.cvssLink.Click += new RoutedEventHandler(delegate (object obj, RoutedEventArgs eventArgs)
-                {
-                    if (obj is Hyperlink)
-                    {
-                        var link = obj as Hyperlink;
+            }
+        }
 
-                        Process.Start(new ProcessStartInfo(link.NavigateUri.AbsoluteUri));
-                    }
-                });
+        private void OnCvssLinkClick(object obj, RoutedEventArgs eventArgs)
+        {
+            if (obj is Hyperlink link)
+            {
+                Process.Start(new ProcessStartInfo(link.NavigateUri.AbsoluteUri));
             }
         }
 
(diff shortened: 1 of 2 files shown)


Snyk.VisualStudio.Extension.Shared/UI/Toolwindow/DescriptionHeaderPanel.xaml.cs
```diff
@@ -137,26 +136,26 @@ private void AddLinksToPanel(StackPanel panel, string[] linkIds, string namePref
...
                 var link = new Hyperlink();
 
-                    link.NavigateUri = new Uri(string.Format(urlPatter, linkData.Replace(namePrefix, string.Empty)));
-                    link.Inlines.Add(linkData);
-                link.Click += new RoutedEventHandler(delegate (object obj, RoutedEventArgs args)
+                link.NavigateUri = new Uri(string.Format(urlPatter, linkData.Replace(namePrefix, string.Empty)));
+                link.Inlines.Add(linkData);
+                link.Click += new RoutedEventHandler(delegate(object obj, RoutedEventArgs args)
+                {
+                    if (obj is Hyperlink)
```
michelkaporin, line 147: We can unify the code here and above on line 85.

Snyk.VisualStudio.Extension.Shared/UI/Toolwindow/DescriptionHeaderPanel.xaml.cs
```diff
@@ -137,26 +136,26 @@ private void AddLinksToPanel(StackPanel panel, string[] linkIds, string namePref
...
+            {
                 var textBlock = new TextBlock();
                 var link = new Hyperlink();
 
-                    link.NavigateUri = new Uri(string.Format(urlPatter, linkData.Replace(namePrefix, string.Empty)));
-                    link.Inlines.Add(linkData);
-                link.Click += new RoutedEventHandler(delegate (object obj, RoutedEventArgs args)
+                link.NavigateUri = new Uri(string.Format(urlPatter, linkData.Replace(namePrefix, string.Empty)));
+                link.Inlines.Add(linkData);
+                link.Click += new RoutedEventHandler(delegate(object obj, RoutedEventArgs args)
```
michelkaporin, line 145: This will probably result in the same issue as you've fixed for CvssLink. This function is called every time Set for Vulnerability is invoked. Can we do the same trick as there but here?

michelkaporin (changes requested):

Snyk.VisualStudio.Extension.Shared/UI/Toolwindow/DescriptionHeaderPanel.xaml.cs
asaf92: In this case it won't happen because it's a `new` HyperLink object (so no previously subscribed event handlers), but I think it's preferable to not use anonymous methods here regardless.

michelkaporin: @asaf92 makes sense, thank you!

michelkaporin:
@asaf92 for the future - we follow angular style [commit](https://github.com/angular/angular.js/blob/master/DEVELOPERS.md#commits) both for branches and commit messages. You can read more about our git methodologies [here](https://www.notion.so/snyk/Git-Work-Methodologies-be6d298d764d47ca998d8f84991f4344#f2af5841f22446f28b3c27aec7a53642).

michelkaporin (approved):
Lovely! Congratulations on your first contribution! 👏 🚀

PR #116, opened 2022-04-05
[ROAD-815] fix: change file size limit

asaf92:
Changing file size limit from 4MB to 1MB


Snyk.Code.Library/Api/SnykCodeClient.cs
```diff
@@ -15,9 +15,9 @@
     public class SnykCodeClient : ISnykCodeClient
     {
         /// <summary>
-        /// Maxium bundle size per one upload is 4 Mb. 4 Mb in bytes.
+        /// Maximum bundle size (in bytes) per one file upload. The maximum size is 1 Mb.
         /// </summary>
-        public const int MaxBundleSize = 4000000;
+        public const int MaxBundleSize = 1_000_000;
```
michelkaporin, line 20: This is a bundle (collection of files) size, not a file one. See the Slack thread for the place I believe change has to happen :) Also a test validating this would be nice to have (if we don't have one already).

michelkaporin (changes requested):

Snyk.Code.Library/Api/SnykCodeClient.cs
asaf92: Thanks, I changed `MaxFileSize` to 1MB and added a test

michelkaporin (approved):
Great stuff 🚀

PR #125, opened 2022-04-12
fix: vsix manifest length exceeding schema limit

michelkaporin:
### Description
Fixes "VSSDK: error VsixPub0024 : The description for the extension must be set and it must be less than 280 characters in length." as reported in [the latest failed release](https://github.com/snyk/snyk-visual-studio-plugin/runs/5990477867?check_suite_focus=true).


aldanchenko (approved):
LGTM

PR #131, opened 2022-04-13
fix: extension id for automated publishing

michelkaporin:
### Description
Attempt to fix `VSSDK: error VsixPub0029 : An error occurred while communicating with the marketplace: vsixId: The VSIX ID in the provided file is already in use. Every extension needs to have an unique ID. Provide a file with an unique ID.`
- Use internal name from Marketplace as Identity Id in VSIX manifest. 
- Revert last PR with publisher name change


aldanchenko (approved):
👍

PR #136, opened 2022-04-15
[ROAD-833] Fix: open source scan issues - fixed root directory for scan calculation

aldanchenko:
### Description

Before this changes root directory for cli scan calculated using solution file. Now extension will get solution file and check is it root directory for solution file and other projects. If not root directory for all projects in solution it will find parent root directory for all projects in solution. 

Because now this method is using async/await changes affected other files.

### Checklist

- ✅ Tests added and all succeed
- ✅ Linted
- ✅ CHANGELOG.md updated
- ☑️ README.md updated, if user-facing

### Screenshots / GIFs

Example of project with solution file not in root directory.
 
Screenshot 2022-04-09 at 13 19 14

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 
 ### Fixed
 - "Object reference not set to an instance of an object" when launching extension in Visual Studio 2022.
+- Scan of solutions in which *.sln file is not in the root directory.
 
 ## [1.1.8]
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,13 +1,17 @@
 # Snyk Changelog
 
-## [1.1.9]
+## [1.1.10]
 
 ### Changed
 - Snyk Code: add support for Single Tenant setups.
 
 ### Fixed
-- "Object reference not set to an instance of an object" when launching extension in Visual Studio 2022.
 - Scan of solutions in which *.sln file is not in the root directory.
+
+## [1.1.9]
+
+### Fixed
+- "Object reference not set to an instance of an object" when launching extension in Visual Studio 2022.
 
 ## [1.1.8]
 
--- a/Snyk.Code.Library/Service/CodeCacheService.cs
+++ b/Snyk.Code.Library/Service/CodeCacheService.cs
@@ -109,7 +109,7 @@
 
                     if (this.filePathToHashCache[filePath] == null)
                     {
-                        await this.AddFileAsync(file);
+                        await this.AddFileAsync(file); // TODO: improve performance, parallelise this work (ROAD-871)
                     }
 
                     filePathToHashDict.Add(filePath, this.filePathToHashCache[filePath].ToString());
--- a/Snyk.Code.Library/Service/CodeCacheService.cs
+++ b/Snyk.Code.Library/Service/CodeCacheService.cs
@@ -5,6 +5,7 @@
     using System.IO;
     using System.Runtime.Caching;
     using System.Text;
+    using System.Threading.Tasks;
     using Serilog;
     using Snyk.Code.Library.Domain.Analysis;
     using Snyk.Common;
@@ -96,7 +97,7 @@
         }
 
         /// <inheritdoc/>
-        public IDictionary<string, string> GetFilePathToHashDictionary(IEnumerable<string> files)
+        public async Task<IDictionary<string, string>> GetFilePathToHashDictionaryAsync(IEnumerable<string> files)
         {
             var filePathToHashDict = new Dictionary<string, string>();
 
@@ -104,11 +105,11 @@
             {
                 if (File.Exists(file))
                 {
-                    string filePath = this.GetRelativeFilePathIfFullPath(file);
+                    string filePath = await this.GetRelativeFilePathIfFullPathAsync(file);
 
                     if (this.filePathToHashCache[filePath] == null)
                     {
-                        this.AddFile(file);
+                        await this.AddFileAsync(file);
                     }
 
                     filePathToHashDict.Add(filePath, this.filePathToHashCache[filePath].ToString());
@@ -141,7 +142,7 @@
         {
             foreach (string filePath in files)
             {
-                this.AddFile(filePath);
+                this.AddFileAsync(filePath);
             }
         }
 
@@ -161,18 +162,20 @@
 
             foreach (string file in fileProvider.GetRemovedFiles())
             {
-                this.RemoveFile(file);
+                this.RemoveFileAsync(file);
             }
         }
 
         /// <inheritdoc/>
-        public IEnumerable<string> GetRelativeFilePaths(IEnumerable<string> files)
+        public async Task<IEnumerable<string>> GetRelativeFilePathsAsync(IEnumerable<string> files)
         {
             IList<string> relateFilePaths = new List<string>();
 
             foreach (string fileFullPath in files)
             {
-                relateFilePaths.Add(FileUtil.GetRelativeFilePath(this.fileProvider.GetSolutionPath(), fileFullPath));
+                var solutionPath = await this.fileProvider.GetSolutionPathAsync();
+
+                relateFilePaths.Add(FileUtil.GetRelativeFilePath(solutionPath, fileFullPath));
             }
 
             return relateFilePaths;
@@ -182,7 +185,7 @@
         /// Add (or update if it already exists) file hash and content to cache.
         /// </summary>
         /// <param name="filePath">Source file path.</param>
-        private void AddFile(string filePath)
+        private async Task<string> AddFileAsync(string filePath)
         {
             try
             {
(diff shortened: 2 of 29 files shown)


Snyk.VisualStudio.Extension.Shared/UI/Toolwindow/SnykCode/DataFlowStepsControl.xaml.cs
```diff
@@ -99,16 +102,8 @@ internal void Display(IList<Marker> markers)
...
-                        {
-                            VsCodeService.Instance.OpenAndNavigate(
-                                this.solutionService.GetFileFullPath(position.FileName),
-                                startLine,
-                                startColumn,
-                                endLine,
-                                endColumn);
-                        })),
+                        LineContent = await this.GetLineContentAsync(position.FileName, startLineNumber),
+                        NavigateCommand = new DelegateCommand((obj) => this.NavigateToCodeAsync(position.FileName, startLine, startColumn, endLine, endColumn)),
```
aldanchenko, line 106: @asaf92 could you please help with this? Don't sure how to change it and avoid void + async usage. Thanks!

asaf92: For now the solution is probably to use the `FireAndForget()` method from the Community VisualStudio Toolkit.
There are async `ICommand` implementations in many MVVM toolkits but we'll talk about that later.

We should probably double check how exceptions are handled before and after this change. `FireAndForget` should prevent crashes, but I'm not sure it would report in Sentry correctly

Snyk.VisualStudio.Extension.Shared/UI/Toolwindow/SnykCode/DataFlowStepsControl.xaml.cs
```diff
@@ -68,7 +70,8 @@ public void AddDataFlowSteps(IList<DataFlowStep> dataFlowSteps)
         /// Add markers to panel.
         /// </summary>
         /// <param name="markers">Markers from suggestion.</param>
-        internal void Display(IList<Marker> markers)
+        /// <returns>A <see cref="Task"/> representing the asynchronous operation.</returns>
+        internal async System.Threading.Tasks.Task DisplayAsync(IList<Marker> markers)
```
asaf92, line 74: I think adding a `using Task = System.Threading.Tasks.Task` at the top looks cleaner than fully qualifying every `Task`

asaf92 (commented):
I love all the async refactoring going on here :)

Snyk.VisualStudio.Extension.Shared/Service/SnykSolutionService.cs
```diff
@@ -127,53 +128,34 @@ public string GetFileFullPath(string file)
         /// If no success, try to get path for flat project (without solution) or web site (in case VS2015).
         /// </summary>
         /// <returns>Solution path string.</returns>
-        public string GetPath()
+        public async System.Threading.Tasks.Task<string> GetPathAsync()
```
michelkaporin, line 131: Let's rename GetPathAsync to `GetSolutionFolder()`.

I was reading the code thinking that path is a full path to the solution file, whereas it is not... It took me a while to figure out what the logic is doing. 🤯

Snyk.VisualStudio.Extension.Shared/Service/SnykSolutionService.cs
```diff
@@ -127,53 +128,34 @@ public string GetFileFullPath(string file)
...
-                {
-                    Logger.Information("Remove solution file name from path.");
+            return rootDir;
+        }
 
-                    solutionPath = Directory.GetParent(solutionPath).FullName;
-                }
+        /// <inheritdoc/>
+        public string FindRootDirectoryForSolutionProjects(string rootDir, IList<string> projectDirs)
+        {
```
michelkaporin, line 145: Interesting solution 👍

Snyk.VisualStudio.Extension.Shared/Service/SnykTasksService.cs
```diff
@@ -368,72 +368,79 @@ private void ScanOss(FeaturesSettings featuresSettings)
...
+            catch (Exception ex)
+            {
+                Logger.Error(ex, "Error on oss scan");
+            }
+        }
 
-                        this.FireOssScanningStartedEvent();
+        private async Task RunOssScanAsync()
+        {
+            {
```
michelkaporin, line 381: Wow, first time I see that subsequent curly braces are allowed in C# 😮 Do we really need it here?

Snyk.VisualStudio.Extension.Shared/Settings/SnykSolutionOptionsUserControl.cs
```diff
@@ -79,42 +80,45 @@ private void SnykProjectOptionsUserControl_Load(object sender, EventArgs eventAr
...
-                this.allProjectsCheckBox.Checked = false;
+                    this.allProjectsCheckBox.Checked = false;
 
-                this.userStorageSettingsService.SaveIsAllProjectsScanEnabled(false);
-            }
+                    await this.userStorageSettingsService.SaveIsAllProjectsScanEnabledAsync(false);
+                }
 
-            this.CheckOptionConflicts();
+                this.CheckOptionConflicts();
```
michelkaporin, line 120: This should be wrapped in try/catch if you run the task like this (fire-and-forget). Don't need to address this comment if not run in `Task.Run()` after removing TODO.

Snyk.VisualStudio.Extension.Shared/Settings/SnykSolutionOptionsUserControl.cs
```diff
@@ -79,42 +80,45 @@ private void SnykProjectOptionsUserControl_Load(object sender, EventArgs eventAr
                 return;
             }
 
-            try
+            Task.Run(async () => // TDOO: Don't sure how to call async/await.
```
michelkaporin, line 83: Do you need help on this? Can [this answer](https://social.msdn.microsoft.com/Forums/sqlserver/en-US/a71a3b04-b19b-4ff6-8f21-2c960c25e397/how-can-i-call-asynchronous-method-in-an-event-handler?forum=csharpgeneral) perhaps help?

CHANGELOG.md
```diff
@@ -7,6 +7,7 @@
 
 ### Fixed
 - "Object reference not set to an instance of an object" when launching extension in Visual Studio 2022.
+- Scan of solutions in which *.sln file is not in the root directory.
```
michelkaporin, line 10: This should be 1.1.10 already. We've release 1.1.9

Snyk.Code.Library/Service/CodeCacheService.cs
```diff
@@ -96,19 +97,19 @@ public class CodeCacheService : ICodeCacheService
         }
 
         /// <inheritdoc/>
-        public IDictionary<string, string> GetFilePathToHashDictionary(IEnumerable<string> files)
+        public async Task<IDictionary<string, string>> GetFilePathToHashDictionaryAsync(IEnumerable<string> files)
         {
             var filePathToHashDict = new Dictionary<string, string>();
 
             foreach (string file in files)
```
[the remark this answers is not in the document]
aldanchenko: Created Jira ticket - ROAD-871 for this

Snyk.VisualStudio.Extension.Shared/Service/SnykTasksService.cs
asaf92: It's allowed (you can limit scope of local variables this way) but probably a mistake. In this case other than the extra indentation it doesn't matter.

This code used to be a parameter for Task.Run, so this is where the curly braces come from.

Snyk.VisualStudio.Extension.Shared/Settings/SnykSolutionOptionsUserControl.cs
asaf92: I think there are 2 better options:
1. Have this method `async void` (an `async` lambda is `async void` anyway) and either `await` this `Task.Run` or avoid it 
2. Use `FireAndForget` 

The code inside the Task.Run has proper exception handling so it doesn't matter too much, but `FireAndForget` helps with some other edge cases.

A more complete solution would be to use MVVM with AsyncCommands

Snyk.VisualStudio.Extension.Shared/Service/SnykSolutionService.cs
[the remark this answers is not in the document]
aldanchenko: Using alias here leads to next error: CS0307 The using alias 'Task' cannot be used with type arguments. I tried to add different alias types, but as for me it's confusing even more:

```
using StringTask = System.Threading.Tasks.Task<string>;
using BoolTask = System.Threading.Tasks.Task<bool>;
```

etc.

what do you think?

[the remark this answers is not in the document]
michelkaporin: it shouldn't be an alias, it should be a using:  
`using System.Threading.Tasks;`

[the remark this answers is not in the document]
aldanchenko: If we use `using System.Threading.Tasks;` it leads to error because `Microsoft.VisualStudio.Shell `also has Task class.

Error message: _CS0104 'Task' is ambiguous reference between 'Microsoft.VisualStudio.Shell.Task' and 'System.Threading.Tasks.Task'._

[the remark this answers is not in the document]
michelkaporin: Got it! Let's keep it prefixed as is then :) Ignore my initial message 👍 Thanks for explaining!

michelkaporin (approved):
LGTM! 👍

asaf92 (approved):

PR #147, opened 2022-05-02
fix: add missing environment parameter

asaf92:
### Description

Fixed secrets missing from the release pipeline


aldanchenko (approved):
LGRM 👍

PR #162, opened 2022-05-25
fix: text color in tree view issues

asaf92:
### Description

Fixed a bug where the color of the text in the tree view did not match the color from VS theme.

### Checklist
- [x] CHANGELOG.md updated


aldanchenko (approved):
LGTM 👍

PR #164, opened 2022-05-27
feat: encode and compress POST/PUT requests to Code API [ROAD-909]

michelkaporin:
### Description

- Encode and compresses POST/PUT requests to Code API [ROAD-909](https://snyksec.atlassian.net/browse/ROAD-90
- Ditch preview manifest files, since they're of no use any more.
- Set empty 0.0.0 version in the manifest, since it's being populated from GitHub tags and value is ambiguous as it is now.

### Checklist

- [x] Linted
- [x] CHANGELOG.md updated


Snyk.VisualStudio.Extension.2022/source.extension.vsixmanifest
```diff
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="utf-8"?>
 <PackageManifest Version="2.0.0" xmlns="http://schemas.microsoft.com/developer/vsx-schema/2011" xmlns:d="http://schemas.microsoft.com/developer/vsx-schema-design/2011">
     <Metadata>
-        <Identity Id="snyk_visual_studio_plugin_2022.1a8ce03c-713d-41f9-9d2d-6451639dfca5" Version="1.1.13" Language="en-US" Publisher="Snyk" />
+        <Identity Id="snyk_visual_studio_plugin_2022.1a8ce03c-713d-41f9-9d2d-6451639dfca5" Version="0.0.0" Language="en-US" Publisher="Snyk" />
```
asaf92, line 4: Why 0.0.0? Shouldn't this be consistent with the actual version?

Also FYI: I noticed that changing this version field can lead to a false error in the IDE saying that the plugin is not installed, but I think a simple rebuild gets around it.

aldanchenko: I think it will be good to discuss it and have one opinion around this. I like Michel approach so we don't need to change it. But if it could lead to unexpected errors maybe better will be to update this property each time

Snyk.Code.Library/Api/SnykCodeClient.cs
```diff
@@ -277,5 +249,30 @@ public string GetAnalysisResultRequestPayload(string bundleId)
                 },
                 Legacy = true,
             });
+
+        private HttpRequestMessage NewHttpRequestMessage(HttpMethod method, string requestUri, HttpContent content)
+        {
+            var request = new HttpRequestMessage(method, requestUri);
+            request.Content = content;
+
+            if (method == HttpMethod.Put || method == HttpMethod.Post)
```
asaf92, line 258: Is this true for all future Put/Post requests made from `SnykCodeClient`?

Snyk.Code.Library/Api/SnykCodeClient.cs
```diff
@@ -212,26 +192,20 @@ public async Task<BundleResponseDto> CreateBundleAsync(IDictionary<string, strin
                 throw new ArgumentException("Bundle files is null.");
             }
 
-            using (var httpRequest = new HttpRequestMessage(HttpMethod.Post, BundleApiUrl))
+            string payload = Json.Serialize(pathToHashFileDict);
+
+            using (var content = await this.NewHttpRequestContentAsync(HttpMethod.Post, payload))
+            using (var httpRequest = this.NewHttpRequestMessage(HttpMethod.Post, BundleApiUrl, content))
+            using (var response = await this.httpClient.SendAsync(httpRequest))
```
asaf92, line 199: You can pass the cancellation token to `SendAsync`

Snyk.Code.Library/Api/SnykCodeClient.cs
```diff
@@ -147,30 +137,24 @@ public async Task<AnalysisResultDto> GetAnalysisAsync(string bundleId, Cancellat
...
-                    Files = pathToHashFileDict,
-                    RemovedFiles = removedFiles,
-                });
+                Files = pathToHashFileDict,
+                RemovedFiles = removedFiles,
+            });
+
+            using (var content = await this.NewHttpRequestContentAsync(HttpMethod.Put, payload))
+            using (var httpRequest = this.NewHttpRequestMessage(HttpMethod.Put, BundleApiUrl + "/" + bundleId, content))
+            using (var response = await this.httpClient.SendAsync(httpRequest))
```
asaf92, line 148: Pass cancellation token to `SendAsync`

Snyk.Code.Library.Tests/SnykCode/Api/SnykCodeClientTest.cs
```diff
@@ -1,4 +1,6 @@
-﻿namespace Snyk.Code.Library.Tests.Api
+﻿using Snyk.Code.Library.Tests.Api;
```
asaf92, line 1: Using directive inside namespace

Snyk.Code.Library.Tests/SnykCode/Api/SnykCodeClientTest.cs
```diff
@@ -22,7 +24,7 @@ public class SnykCodeClientTest
 
         public SnykCodeClientTest()
         {
-            this.snykCodeClient = new SnykCodeClient(
```
asaf92, line 25: Was anything changed in this file other than style? Because if not let's revert it

asaf92 (commented):
- Can the version field be safely set to 0.0.0? 
- Pass cancellation tokens to `SendAsync`
- Add a short comment that explains why the content type/encoding is the way it is on Put/Post requests
- Revert code style changes in that test file (for now 😄)

Snyk.Code.Library/Api/SnykCodeClient.cs
michelkaporin: Yes, that's the API expectation now.

Snyk.VisualStudio.Extension.2022/source.extension.vsixmanifest
michelkaporin: @asaf92 @aldanchenko I'll keep it as-is. Just tested with @asaf92 and it seems we can run it in debug, if it's set to 0.0.0. If this causes problems, try reseting experimental instance. If still causes problems, we can change it to 0.0.1 :)

Snyk.Code.Library/Api/SnykCodeClient.cs
michelkaporin: done

Snyk.Code.Library/Api/SnykCodeClient.cs
michelkaporin: done 👍

Snyk.Code.Library.Tests/SnykCode/Api/SnykCodeClientTest.cs
michelkaporin: yep, fixed

Snyk.Code.Library.Tests/SnykCode/Api/SnykCodeClientTest.cs
michelkaporin: I've reverted, some small tweaks left

asaf92 (dismissed):
LGTM 🚀

asaf92 (approved):

PR #169, opened 2022-05-31
Fix/inactive selected tree item color

asaf92:
### Description

Fixed selection of tree view items only working when clicking on the icon, and an issue where background color of unfocused selected items might blend with font color on some themes

### Checklist
- [ ] CHANGELOG.md updated


aldanchenko (approved):
Cool 🚀

PR #171, opened 2022-06-02
feat: add plugin version to log

asaf92:
### Description

Add plugin version to the log


Snyk.VisualStudio.Extension.Shared/Service/SnykService.cs
```diff
@@ -319,5 +324,31 @@ private void InitializeAnalyticsService()
             this.analyticsService = SnykAnalyticsService.Instance;
             Logger.Information("Analytics service initialized");
         }
+
+        private string GetVsixVersion()
```
michelkaporin, line 328: - Have you considered baking this into the `AssemblyName.Version` instead of reading it from file system? 
- Should this function leave here or rather go to a class that's responsible for returning Visual Studio version in runtime, maybe some other meta settings of the plugin?

PR #184, opened 2022-06-27
[ROAD-719] test: Add integration tests

asaf92:
### Description

Added VS 2019 & VS 2022 integration tests

### Checklist

### Screenshots / GIFs

image


asaf92:
Checking if tests are discoverable & green in the CI. Might need to bump up the MSBuild version

.github/workflows/release.yml
```diff
@@ -47,7 +47,58 @@ jobs:
...
+      run: |
+        msbuild ${{env.SOLUTION_FILE_PATH}} /p:configuration=Release /p:DeployExtension=false /p:ZipPackageCompressionLevel=normal /v:m
+      shell: powershell
+
+    - name: Tests
+      run: vstest.console.exe **\*.Tests.dll
+
+  build-and-release:
+    needs: build-2019
+    runs-on: windows-2022
```
michelkaporin, line 68: Like that we are explicit about version now 👍

.github/workflows/release.yml
```diff
@@ -13,8 +13,8 @@ env:
   DEFAULT_BRANCH: main
 
 jobs:
-  build:
-    runs-on: windows-latest
+  build-2019:
```
michelkaporin, line 16: Should this be called a test job? How about running integration tests against different VS versions? 2015, 2017, etc.

.github/workflows/release.yml
```diff
@@ -47,7 +47,58 @@ jobs:
     - name: Add MSBuild to PATH
```
michelkaporin, line 47: Line 38: do we still need variable substitution for testing?

Tests/SnykVsIntegrationTests.Shared/ExtensionStartupTests.cs
```diff
@@ -0,0 +1,70 @@
...
+			Assert.True(uiShell != null, "Failed to load UI shell");
+
+			// Act
+			this.output.WriteLine("Opening Snyk tool-window");
+			uiShell.PostExecCommand(SnykGuids.SnykVSPackageCommandSet, 
+				(uint) SnykGuids.OpenToolWindowCommandId, 
+				0,
+				null);
+
+			// It's necessary to wait here because when the test finishes too fast, the "Xunit.Instances.VisuaStudio (...)"
```
michelkaporin, line 49: Can we have teardown test suite hook to shut down the "Xunit.Instances.VisuaStudio (...)"?

Tests/SnykVsIntegrationTests.Shared/ExtensionStartupTests.cs
```diff
@@ -0,0 +1,70 @@
...
+			var guid = Guid.Parse(SnykVSPackage.PackageGuidString);
+			var packageObject = await shell.LoadPackageAsync(ref guid);
+			
+			// Assert
+			Assert.True(packageObject != null, "Failed to find Snyk package");
+			Assert.IsType<SnykVSPackage>(packageObject);
+			var snykVsPackage = packageObject as SnykVSPackage;
+			Assert.True(snykVsPackage.ToolWindow != null, "Tool window failed to open");
+			Assert.True(snykVsPackage.IsInitialized, "Snyk package was not initialized");
+			this.output.WriteLine("Test complete!");
```
michelkaporin, line 67: Does this output serve any purpose when executed as last?

Tests/SnykVsIntegrationTests.Shared/GlobalSuppressions.cs
```diff
@@ -0,0 +1,12 @@
+﻿// This file is used by Code Analysis to maintain SuppressMessage
+// attributes that are applied to this project.
+// Project-level suppressions either have no target or are given
+// a specific target and scoped to a namespace, type, member, etc.
+
+using System.Diagnostics.CodeAnalysis;
+
+[assembly: SuppressMessage("Style",
```
michelkaporin, line 8: Can't VSTHRD200 be added to project level supressions? I remember that it used to be on .csproj level in other projects I've worked with.

Tests/Integration.Pre22.Tests/Integration.Pre22.Tests.csproj
```diff
@@ -0,0 +1,65 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
```
michelkaporin, line 1: Can you please explain why Pre22.Tests and Integration.Tests projects are required? Why Shared project alone is not enough?

Tests/SnykVsIntegrationTests.Shared/GlobalSuppressions.cs
asaf92: This is a shared project so there's no `.csproj` file
I'm surprised it still works because the `Target` should have the project name which was changed since I added this suppression.

Tests/SnykVsIntegrationTests.Shared/ExtensionStartupTests.cs
asaf92: Nope, just wanted to check that the test actually finishes when I experimented with it. I'll remove it.

Tests/SnykVsIntegrationTests.Shared/ExtensionStartupTests.cs
michelkaporin: If it's not done, can we programmatically enforce it? Waiting is usually more prone to problems than forcing something programmatically.

asaf92: I agree in general, but this is an integration test scenario. The GH action kills any orphaned processes so it will close the IDE process anyway.

michelkaporin: Don't spend too much time on it tho, timebox it and let me know if no luck.

asaf92: Added an awaiter for package initialization

.github/workflows/build.yml
```diff
@@ -49,4 +49,4 @@ jobs:
       shell: powershell
 
     - name: Tests
-      run: vstest.console.exe **\*.Tests.dll
\ No newline at end of file
+      run: vstest.console.exe **\*.Tests.dll /TestCaseFilter:"FullyQualifiedName!=Xunit.Instances.VisualStudio&integration!=true" #exclude integration tests and the psuedo-tests that launch a VS instance
```
michelkaporin, line 52: What are the pseudo-tests?

asaf92: In the test explorer you'll see a test called `Xunit.Instances.VisualStudio (2019/2022)`. It launches the VS experimental instance.

I want to skip them because filtering the integration tests themselves is not enough, and you'd still have a "test" that simply boots up a VS instance for no reason.

michelkaporin: Interesting, thanks for explaining!

snyk-visual-studio-plugin.sln
```diff
@@ -50,6 +51,14 @@ Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Analytics", "Analytics", "{
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Itly", "Snyk.Analytics.Itly\Itly\Itly.csproj", "{AD9ECE9E-DEF1-4976-998F-D124737994E8}"
 EndProject
+Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Tests", "Tests", "{ADBE502B-B340-41A1-AC16-41E00AE86D55}"
+EndProject
+Project("{D954291E-2A0B-460D-934E-DC6B0785DB48}") = "SnykVsIntegrationTests.Shared", "Tests\SnykVsIntegrationTests.Shared\SnykVsIntegrationTests.Shared.shproj", "{C2E8C32B-BD9F-4927-B6EA-BF6E9B40CB6C}"
```
michelkaporin, line 56: Didn't see it but let's ditch "Snyk" prefix. This has been discussed many times in team that we should avoid it unless really necessary. It doesn't provide any semantic value.

michelkaporin (approved):

michelkaporin:
Amazing stuff, let's not forget to maintain this and keep covering most important plugin flows here 🚀

PR #186, opened 2022-06-28
fix: broken images in readme sync

michelkaporin:
### Description

This fixes broken images when syncing readme from Gitbook.

Marketplace Markdown doesn't support special characters `<>` when adding a Markdown image. E.g. the following Markdown will not be printed correctly and instead interpreted as plaintext:

`Visual Studio Code extension results (1) (1) (1) (1) (1) (1) (1) (1) (1) (1) (1) (1) (1) (1) (1) (1) (1) (1) (1) (1) (1) (1) (1) (1).png>)`.

The fix is to replace these with  HTML tag using `sed` as part of readme sync. 

### Checklist

- [ ] Linted
- [ ] CHANGELOG.md updated

### DIFF
--- a/.github/workflows/readme-sync.yml
+++ b/.github/workflows/readme-sync.yml
@@ -24,7 +24,7 @@ jobs:
               ./snyk-visual-studio-plugin/README.md
           sed -i \
               -E "s|^\!\\[([[:alnum:][:space:][:punct:]]+)\]\(<([[:alnum:][:punct:]\-\.\/:[:space:]\(\)]+)>\)|<img src=\"\2\" alt=\"\1\" />|g" \
-              ./vscode-extension/README.md
+              ./snyk-visual-studio-plugin/README.md
           sed -i \
               -E 's|(\{%.*%\})||g' \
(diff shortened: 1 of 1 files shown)


.github/workflows/readme-sync.yml
```diff
@@ -22,6 +22,9 @@ jobs:
           sed -i \
               -e "s|../.gitbook/assets/|https://github.com/snyk/user-docs/raw/HEAD/docs/.gitbook/assets/|g" \
               ./snyk-visual-studio-plugin/README.md
+          sed -i \
+              -E "s|^\!\\[([[:alnum:][:space:][:punct:]]+)\]\(<([[:alnum:][:punct:]\-\.\/:[:space:]\(\)]+)>\)|<img src=\"\2\" alt=\"\1\" />|g" \
+              ./vscode-extension/README.md
```
asaf92, line 27: `vscode`?

asaf92 (changes requested):

asaf92 (approved):

PR #188, opened 2022-06-29
[ROAD-1017] Fix: Custom endpoint settings

asaf92:
### Description

Fixed custom endpoint changing leading to authentication errors

### Checklist
- [x] CHANGELOG.md update


Snyk.VisualStudio.Extension.Shared/Settings/SnykGeneralOptionsDialogPage.cs
```diff
@@ -66,6 +66,11 @@ public string CustomEndpoint
                     return;
                 }
 
+                // When changing the API endpoint, the API token is invalidated
```
michelkaporin, line 69: How about adding tests for this?

PR #207, opened 2022-08-23
[ROAD-1111] Encryption error workaround

asaf92:
### Description

Attempt to workaround rare "only one reader is allowed" errors when trying to encrypt and encode Snyk Code messages.

### Checklist

- [ ] CHANGELOG.md updated


CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 ## [1.1.26]
 
+### Fixed
+- Rare bug in Code scans encoding.
```
michelkaporin, line 6: Let's be more explicit about the bug. The sentence is rather ambiguous.

Snyk.Code.Library/Api/SnykCodeClient.cs
```diff
@@ -269,10 +269,15 @@ private async Task<HttpContent> NewHttpRequestContentAsync(HttpMethod method, st
             // Snyk Code PUT and POST requests must be base64 encoded and deflated for certain environments (ROAD-909)
             if (method == HttpMethod.Put || method == HttpMethod.Post)
             {
+                Logger.Information("Encoding and compressing {Length} bytes...", System.Text.Encoding.UTF8.GetByteCount(payload));
```
michelkaporin, line 272: Why do we need this information in log?

Snyk.VisualStudio.Extension.Shared/Service/SnykTasksService.cs
```diff
@@ -662,7 +662,7 @@ private IList<AnalysisType> GetSelectedFeatures(FeaturesSettings featuresSetting
             return selectedProducts;
         }
 
-        private void CancelTask(CancellationTokenSource tokenSource)
+        private void CancelTask(ref CancellationTokenSource tokenSource)
```
michelkaporin, line 665: Do you need to pass by reference because we do `tokenSource = null` inside the method? If yes, pardon my C#, doesn't it set the field to null without it? 🤔

asaf92: Exactly. Object references are normally passed by value.

Snyk.Code.Library/Api/SnykCodeClient.cs
asaf92: I want to check whether the errors are related to the size of the chunks we're encoding

michelkaporin (approved):

PR #229, opened 2023-05-04
fix: integration tests workflow [HEAD-294]

asaf92:
### Description

Fix the release workflow. Build on VS22 and run on VS19.

This change:
- Renames "build.yml" to "pr-workflow.yml"
- Added a "build-project.yml" reusable workflow
- Added a check that verifies that the secrets are accessible in the build step
- Added an "integration-tests.yml" reusable workflow
- Modified both the pr-workflow and the release workflow to use the reusable workflows


bastiandoetsch (approved):
LGTM, please fix the branch name :)

PR #232, opened 2023-05-31
fix: version in release workflow is wrong for marketplace

asaf92:
### Description

Fixed the wrong version in the marketplace deployment


Snyk.VisualStudio.Extension.Shared/UI/Notifications/NotificationService.cs
```diff
@@ -1,12 +1,14 @@
...
 {
+    using Snyk.Common;
     using Snyk.VisualStudio.Extension.Shared.Service;
 
     /// <summary>
     /// Snyk VS notification service.
     /// </summary>
     public class NotificationService
     {
+        private const int MaxErrorLength = 300;
```
PeterSchafer, line 11: question: do we have a source for this number? I would expect the UI to limit overflowing text itself.

PeterSchafer (approved):

PR #237, opened 2023-06-18
Chore: add user agent headers [head-373]

asaf92:


Snyk.VisualStudio.Extension.Shared/CLI/SnykCli.cs
```diff
@@ -34,9 +34,9 @@ public class SnykCli : ICli
         /// <summary>
         /// Initializes a new instance of the <see cref="SnykCli"/> class.
         /// </summary>
-        public SnykCli(ISnykOptions options)
+        public SnykCli(ISnykOptions options, string ideVersion = "")
         {
-            this.ConsoleRunner = new SnykConsoleRunner();
+            this.ConsoleRunner = new SnykConsoleRunner(ideVersion);
```
PeterSchafer, line 39: Issue: It appears that the IDE information is not part of the UA header, this is an example of what we receive
`snyk-cli/1.1182.0 (windows;amd64;snyk-win.exe) VISUAL_STUDIO/0.0.0`

asaf92: It's the plugin version, and locally it's set to 0.0.0 so this is normal.

It should look like this though:
`snyk-cli/1.1182.0 (windows;amd64;snyk-win.exe) VISUAL_STUDIO/0.0.0 (VISUAL_STUDIO/<VS_VERSION>)`

I'll check why it's omitting the final part

PR #244, opened 2023-07-27
fix: Custom Endpoint URL crash [HEAD-536]

asaf92:
### Description

Custom endpoints validation in the settings page did not work correctly and led to crashes.

### DIFF
--- a/Snyk.VisualStudio.Extension.Shared/CLI/SnykCli.cs
+++ b/Snyk.VisualStudio.Extension.Shared/CLI/SnykCli.cs
@@ -81,7 +81,7 @@
             }
             catch (InvalidTokenException e)
             {
-                Logger.Warning(e, "Error on get api token via cli for settings");
+                Logger.Warning(e, "Failed to retrieve api token from CLI configuration");
 
                 apiToken = string.Empty;
             }
--- a/Snyk.VisualStudio.Extension.Shared/CLI/SnykCli.cs
+++ b/Snyk.VisualStudio.Extension.Shared/CLI/SnykCli.cs
@@ -81,7 +81,7 @@
             }
             catch (InvalidTokenException e)
             {
-                Logger.Error(e, "Error on get api token via cli for settings");
+                Logger.Warning(e, "Error on get api token via cli for settings");
 
                 apiToken = string.Empty;
             }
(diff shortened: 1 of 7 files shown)


Snyk.VisualStudio.Extension.Shared/CLI/SnykCli.cs
```diff
@@ -81,7 +81,7 @@ public string GetApiToken()
             }
             catch (InvalidTokenException e)
             {
-                Logger.Error(e, "Error on get api token via cli for settings");
+                Logger.Warning(e, "Error on get api token via cli for settings");
```
bastiandoetsch, line 84: `Error getting...` would probably be better.

Snyk.VisualStudio.Extension.Shared/Settings/SnykGeneralOptionsDialogPage.cs
```diff
@@ -93,6 +93,12 @@ public string CustomEndpoint
             get => this.customEndpoint;
             set
             {
+                if (!Uri.IsWellFormedUriString(value, UriKind.Absolute))
```
bastiandoetsch, line 96: Should we add the same endpoint logic as in IntelliJ? https://github.com/snyk/snyk-intellij-plugin/blob/c56af2012a2974f7f52172c0a02effbd2a334523/src/main/kotlin/io/snyk/plugin/Utils.kt#L151

```
fun URI.isSnykTenant() =
    isSnykDomain() && (host.startsWith("app.") || host == "snyk.io" || isDev()) && path.endsWith("/api")
```

And then, when used for the API, the URL is transformed to an API URL:
```
fun URI.toSnykAPIv1(): URI {
    val host = host
        .replaceFirst("app.", "api.")
        .replaceFirst("deeproxy.", "api.")
        .prefixIfNot("api.")

    return URI(scheme, host, "/v1/", null)
}
```

asaf92: It looks like there are already places doing this conversion (`GetAppCustomEndpoint` for example)

bastiandoetsch (approved):
Thanks. Looks good! 🎉

PR #245, opened 2023-08-29
chore: disable analytics & sentry when fedramp [HEAD-677]

bastiandoetsch:
### Description

### Checklist

- [ ] Linted
- [ ] CHANGELOG.md updated

### Screenshots / GIFs

### DIFF
--- a/Snyk.Common/Settings/ISnykOptions.cs
+++ b/Snyk.Common/Settings/ISnykOptions.cs
@@ -14,6 +14,8 @@
         /// </summary>
         AuthenticationToken ApiToken { get; }
 
+        bool IsFedramp();
+
         /// <summary>
         /// Gets or sets a value indicating whether CLI custom endpoint parameter.
         /// </summary>
--- a/Snyk.VisualStudio.Extension.Shared/CLI/SnykConsoleRunner.cs
+++ b/Snyk.VisualStudio.Extension.Shared/CLI/SnykConsoleRunner.cs
@@ -1,4 +1,6 @@
-﻿namespace Snyk.VisualStudio.Extension.Shared.CLI
+﻿using Snyk.Common.Settings;
+
+namespace Snyk.VisualStudio.Extension.Shared.CLI
 {
     using System;
     using System.Collections;
@@ -17,10 +19,12 @@
 
         private bool isStopped = false;
         private readonly string ideVersion;
+        private readonly ISnykOptions options;
 
-        public SnykConsoleRunner(string ideVersion = "")
+        public SnykConsoleRunner(ISnykOptions options, string ideVersion = "")
         {
             this.ideVersion = ideVersion;
+            this.options = options;
         }
 
         /// <summary>
@@ -74,6 +78,11 @@
                 }
             }
 
+            if (!this.options.UsageAnalyticsEnabled || this.options.IsFedramp())
+            {
+                processStartInfo.EnvironmentVariables["SNYK_CFG_DISABLE_ANALYTICS"] = "1";
+            }
+            
             processStartInfo.EnvironmentVariables["SNYK_INTEGRATION_NAME"] = SnykExtension.IntegrationName;
             processStartInfo.EnvironmentVariables["SNYK_INTEGRATION_VERSION"] = SnykExtension.Version;
             processStartInfo.EnvironmentVariables["SNYK_INTEGRATION_ENVIRONMENT_NAME"] = SnykExtension.IntegrationName;
--- a/Snyk.VisualStudio.Extension.Shared/Service/SentryService.cs
+++ b/Snyk.VisualStudio.Extension.Shared/Service/SentryService.cs
@@ -34,17 +34,34 @@
                 scope.SetTag("vs.version", vsVersion.ToString());
                 scope.SetTag("vs.edition", this.serviceProvider.DTE.Edition);
             });
+
+            // disable sentry if no usage analytics or fedramp 
+            if (!this.serviceProvider.Options.UsageAnalyticsEnabled || this.serviceProvider.Options.IsFedramp())
+            {
+                DiscardEventCallback();
+            }
         }
 
         /// <inheritdoc/>
         public void SetSolutionType(SolutionType solutionType)
         {
-            LogManager.SentryConfiguration.BeforeSend = sentryEvent =>
+            if (!this.serviceProvider.Options.UsageAnalyticsEnabled || this.serviceProvider.Options.IsFedramp())
             {
-                sentryEvent.SetTag("vs.project.type", solutionType.ToString());
+                DiscardEventCallback();
+            }
+            else
+            {
+                LogManager.SentryConfiguration.BeforeSend = sentryEvent =>
+                {
+                    sentryEvent.SetTag("vs.project.type", solutionType.ToString());
+                    return sentryEvent;
+                };
+            }
+        }
 
-                return sentryEvent;
-            };
+        private static void DiscardEventCallback()
+        {
+            LogManager.SentryConfiguration.BeforeSend = _ => { return null; };
         }
     }
 }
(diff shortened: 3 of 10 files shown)


Snyk.VisualStudio.Extension.Shared/CLI/SnykConsoleRunner.cs
```diff
@@ -74,6 +78,11 @@ public virtual Process CreateProcess(string fileName, string arguments, StringDi
                 }
             }
 
+            if (!this.options.UsageAnalyticsEnabled || this.options.IsFedramp())
```
PeterSchafer, line 81: Question: Why don't we just extend the getter for UsageAnalyticsEnabled to return different values depending on IsFedramp().

bastiandoetsch: Because I didn't want to mix the UsageAnalyticsSetting with Fedramp. When we move to a different analytics/logging provider or mechanism we just have to remove the check for isFedramp(). It's semantically different, so I didn't integrate it. Thought about it, though.

PeterSchafer: Thank you for sharing your thoughts! I think I don't completely subscribe to them as it feels a bit like a premature optimization. But taking this view, I was wondering if [this line](https://github.com/snyk/snyk-visual-studio-plugin/blob/e3c60f930ae126f803b6b32011c1411f64611798/Snyk.VisualStudio.Extension.Shared/UI/Toolwindow/SnykToolWindowControl.xaml.cs#L690) was missed, shouldn't it contain the additional check as well?

Snyk.VisualStudio.Extension.Shared/Service/SentryService.cs
```diff
@@ -34,17 +34,34 @@ public async Task SetupAsync()
...
+                    return sentryEvent;
+                };
+            }
+        }
 
-                return sentryEvent;
-            };
+        private static void DiscardEventCallback()
+        {
+            LogManager.SentryConfiguration.BeforeSend = _ => { return null; };
```
PeterSchafer, line 64: Question: why is this now necessary as it wasn't there before.

Snyk.VisualStudio.Extension.Tests/SnykCliTest.cs
```diff
@@ -177,28 +180,16 @@ public async Task SnykCliTest_BuildArguments_WithAllOptionsAsync()
...
-                   .Setup(options => options.IsScanAllProjectsAsync())
-                   .ReturnsAsync(true);
+                .Setup(options => options.IsScanAllProjectsAsync())
+                .ReturnsAsync(true);
 
             var cli = new SnykCli(this.optionsMock.Object);
 
             Assert.Equal(
-                "--json test --insecure --org=test-snyk-organization --ignore-policy --all-projects --DISABLE_ANALYTICS",
+                "--json test --insecure --org=test-snyk-organization --ignore-policy --all-projects",
```
PeterSchafer, line 189: Suggestion: There should be an additional check for the changed logic. Ensuring that the Environment Variable is set.

Snyk.VisualStudio.Extension.Shared/Settings/SnykGeneralOptionsDialogPage.cs
```diff
@@ -84,6 +84,22 @@ private AuthenticationToken CreateAuthenticationToken(string token)
...
+        public bool IsFedramp()
+        {
+            var endpoint = this.customEndpoint;
+            if (endpoint.IsNullOrEmpty())
+            {
+                return false;
+            }
+
+            var endpointUri = new Uri(endpoint);
+            return endpoint.ToLower().Contains("fedramp") && endpointUri.Host.EndsWith("snykgov.io");
```
PeterSchafer, line 101: Question: Are we sure that these two checks needs to use logical and? I think only the latter should be correct.

Snyk.VisualStudio.Extension.Shared/CLI/SnykConsoleRunner.cs
bastiandoetsch: Added :)

Snyk.VisualStudio.Extension.Shared/Service/SentryService.cs
bastiandoetsch: It was a bug. Sentry was not respecting the analytics setting.

Snyk.VisualStudio.Extension.Shared/Settings/SnykGeneralOptionsDialogPage.cs
bastiandoetsch: I can remove the fedramp part as discussed

Snyk.VisualStudio.Extension.Shared/Settings/SnykGeneralOptionsDialogPage.cs
```diff
@@ -104,6 +120,11 @@ public string CustomEndpoint
                     return;
                 }
 
+                if (IsFedramp())
+                {
+                    this.UsageAnalyticsEnabled = false;
```
PeterSchafer, line 125: Question: not sure if this is wanted, as it seems to change the settings value.

PeterSchafer: also the order here is a bit tricky, IsFedramp() uses the previous value not the value it is about to set.

PeterSchafer (approved):

Snyk.Common/Settings/ISnykOptions.cs
```diff
@@ -14,6 +14,8 @@ public interface ISnykOptions
         /// </summary>
         AuthenticationToken ApiToken { get; }
 
+        bool IsFedramp();
```
asaf92, line 17: Usually when you have `FooType GetFoo()`, it's idiomatic to use a getter only auto-property, so `FooType Foo { get; }`

Snyk.VisualStudio.Extension.Shared/CLI/SnykConsoleRunner.cs
asaf92: On the one hand I think it's better to use `IsFedramp()` in the `UsageAnalyticsEnabled` getter. The alternative is to remember to check `IsFedramp` everywhere which is somewhat fragile. If we change the logic of when analytics is allowed/disabled, it's better to change it in a single place. 

`options.UsageAnalyticsEnabled` answers the question "Should/can analytics be used", it doesn't have to be a one-to-one mapping to the user settings. We can rename it to `AnalyticsEnabled` or something.

On the other hand, it would require some work on the settings page that I wouldn't want to creep into this PR, because this property also has a setter.

Snyk.VisualStudio.Extension.Shared/Settings/SnykGeneralOptionsDialogPage.cs
```diff
@@ -84,6 +84,22 @@ private AuthenticationToken CreateAuthenticationToken(string token)
...
+        {
+            var endpoint = this.customEndpoint;
+            if (endpoint.IsNullOrEmpty())
+            {
+                return false;
+            }
+
+            var endpointUri = new Uri(endpoint);
+            return endpointUri.Host.ToLower().EndsWith("snykgov.io");
+        }
```
asaf92, line 102: (Suggestion) 
You can also do something like this:
```suggestion
        public bool IsFedramp() => > Uri.TryCreate(this.customEndpoint, UriKind.Absolute, out var uri) &&
                                 uri.Host.ToLower().EndsWith("snykgov.io");
```

It's shorter and handles any error in the URI parsing (like invalid format). I recently added code to the `CustomEndpoint` setter to avoid invalid URIs, but it's better to be safe anyway.

asaf92 (approved):
Left some suggestions

CALISOULB (approved):

PR #257, opened 2024-03-25
fix: upgrading package version on Newtonsoft IDE-218

acke:
### Description

This patch Upgrade the version of the Newtonsoft dependency.

### Checklist

- [x] Linted
- [x] CHANGELOG.md updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Snyk Security Changelog
 
+## [1.1.48]
+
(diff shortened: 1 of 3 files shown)


acke:
For the testing of this solution to this ticket, we have run scans with the new and old versions of this extension.
The same issue results have been reached in both cases. 

Issues found after update
image


Issues found before update
image

cat2608 (approved):

PR #258, opened 2024-03-27
fix: upgrade analytics dependency

acke:
### Description

### Checklist

- [x] Linted
- [x] CHANGELOG.md updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,16 +1,21 @@
 # Snyk Security Changelog
 
-## [1.1.48]
+## [1.1.50]
+
+### Fixed
+- Upgraded the version of the Analytics dependency.
+
+## [1.1.49]
 
 ### Fixed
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -1,16 +1,21 @@
...
+### Fixed
+- Upgraded the version of the Analytics dependency.
+
+## [1.1.49]
 
 ### Fixed
 - Upgraded the version of Newtonsoft dependency.
 
-## [1.1.47]
+## [1.1.48]
```
teodora-sandu, line 13: Why did we change these versions? Based on `main` I would have just expected a new `1.1.49` version to be added for this upgrade: https://github.com/snyk/snyk-visual-studio-plugin/blob/main/CHANGELOG.md

teodora-sandu (approved):
Looks okay to me, I asked a question but it could be me not understanding how this works 😄

PR #269, opened 2024-06-04
fix: resource loading [IDE-385]

ShawkyZ:
### Description

- Determine base path for resources by trying to load the resources with and without "Resources" in the Pack URI.
- Fix a bug when the extension loads and the user is not authenticated

### Checklist

- [ ] Linted
- [x] CHANGELOG.md updated

### Screenshots / GIFs


Snyk.VisualStudio.Extension.Shared/UI/ResourceLoader.cs
```diff
@@ -0,0 +1,52 @@
+using System;
+using System.Windows;
+
+namespace Snyk.VisualStudio.Extension
+{
+    public static class ResourceLoader
```
cmars, line 6: Is it possible to test this class? Not sure how the pack:// uri is resolved but maybe there is a way?

ShawkyZ: We should be able to test the class behavior if a resource exists/doesn't exist

bastiandoetsch (approved):
Nice clean up

PR #288, opened 2024-07-30
feat: OAuth2 option [IDE-461] [IDE-465]

ShawkyZ:
### Description
- Added a new Dropdown for Authentication Method.
- Explicitly determine Auth method.
- Refactor refresh logic to explicitly refresh the token (see changes in: AuthenticationToken.cs).
- Fix fedramp code endpoint.

image

### Checklist

- [x] Linted
- [x] CHANGELOG.md updated

### Screenshots / GIFs


Snyk.Common/HttpClientFactory.cs
```diff
@@ -37,7 +37,7 @@ public static HttpClient NewHttpClient(AuthenticationToken token, string baseUrl
             var authorizationString = "token " + token;
             if (token.Type == AuthenticationType.OAuth)
             {
-                var rawToken = token.ToString();
+                var rawToken = token.Refresh();
```
bastiandoetsch, line 40: As we move to LS, this is ok for now, but in general having a second CLI call for each normal call is sub-optimal :)

bastiandoetsch (approved):

Snyk.Common/HttpClientFactory.cs
ShawkyZ: This is the same call. I extracted the refresh logic in the ToString override to a Refresh method

Snyk.Common.Tests/Service/ApiEndpointResolverTest.cs
```diff
@@ -51,17 +52,34 @@ public void ApiEndpointResolver_GetSnykCodeApiUrl_SingleTenant()
...
+        public void ApiEndpointResolver_GetSnykCodeApiUrl_Snykgov_NoOrg()
         {
             var optionsMock = new Mock<ISnykOptions>();
             optionsMock
                  .Setup(options => options.CustomEndpoint)
                  .Returns("https://app.random-uuid.polaris.snykgov.io/api");
+            
+            var apiEndpointResolver = new ApiEndpointResolver(optionsMock.Object);
+            
+            Assert.Throws<InvalidOperationException>(() => apiEndpointResolver.GetSnykCodeApiUrl());
```
bastiandoetsch, line 67: why? what is wrong?

PR #318, opened 2024-10-24
feat: persist api url from LS after auth

ShawkyZ:
### Description

### Checklist

- [ ] Linted
- [ ] CHANGELOG.md updated

### Screenshots / GIFs

### DIFF
--- a/Snyk.VisualStudio.Extension.2022/Language/SnykLanguageClientCustomTarget.cs
+++ b/Snyk.VisualStudio.Extension.2022/Language/SnykLanguageClientCustomTarget.cs
@@ -7,6 +7,7 @@ using Newtonsoft.Json.Linq;
 using Snyk.Common.Authentication;
 using Snyk.VisualStudio.Extension.Extension;
 using Snyk.VisualStudio.Extension.Service;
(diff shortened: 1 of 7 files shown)


Snyk.VisualStudio.Extension.2022/Language/SnykLanguageClientCustomTarget.cs
```diff
@@ -120,10 +120,16 @@ public async Task OnHasAuthenticated(JToken arg)
             {
                 return;
             }
+            var apiUrl = arg["apiUrl"].ToString();
+            if (string.IsNullOrEmpty(apiUrl))
+            {
+                return;
+            }
```
bastiandoetsch, line 127: I would add a notification (balloon/banner) that the endpoint has been updated from the server.

bastiandoetsch (approved):

Snyk.VisualStudio.Extension.2022/Language/SnykLanguageClientCustomTarget.cs
ShawkyZ: Also there is a bug here.
Endpoint property should be updated before the token (since it currently sets the token to empty).

PR #338, opened 2024-12-16
fix: after vs load init

ShawkyZ:
### Description

### Checklist

- [ ] Linted
- [ ] CHANGELOG.md updated

### Screenshots / GIFs

### DIFF
--- a/Snyk.VisualStudio.Extension.2022/SnykVSPackage.cs
+++ b/Snyk.VisualStudio.Extension.2022/SnykVSPackage.cs
@@ -65,7 +65,7 @@ namespace Snyk.VisualStudio.Extension
             new TaskCompletionSource<bool>();
 
         public static SnykVSPackage Instance;
-        private readonly SemaphoreSlim semaphore = new SemaphoreSlim(1, 1);
+        private readonly SemaphoreSlim languageClientInitSemaphore = new SemaphoreSlim(1, 1);
 
         private ISnykServiceProvider serviceProvider;
 
@@ -261,7 +261,7 @@ namespace Snyk.VisualStudio.Extension
         private Window tempOpenedFileWindow;
         private async Task LanguageClientManagerOnLanguageClientNotInitializedAsync(object sender, SnykLanguageServerEventArgs args)
         {
-            await semaphore.WaitAsync(DisposalToken);
+            await languageClientInitSemaphore.WaitAsync(DisposalToken);
 
             ThreadHelper.JoinableTaskFactory.RunAsync(async () =>
             {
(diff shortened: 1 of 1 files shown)


Snyk.VisualStudio.Extension.2022/SnykVSPackage.cs
```diff
@@ -257,21 +259,47 @@ private async Task LanguageClientManagerOnOnLanguageServerReadyAsync(object send
...
+
+            ThreadHelper.JoinableTaskFactory.RunAsync(async () =>
+            {
+                try
+                {
+                    await JoinableTaskFactory.SwitchToMainThreadAsync();
+                    while (!LanguageClientHelper.IsLanguageServerReady())
+                    {
+                        var isSolutionOrFolderOpen = SnykSolutionService.Instance.IsSolutionOpen();
+                        if (isSolutionOrFolderOpen)
```
andrewrobinsonhodges-snyk, line 274: Rather than have the sleep on line 291 below, I was wondering whether we could instead ensure that this code block only executes once, e.g:

```
bool haveBegunLoadingExtension = false;
if ( !haveBegunLoadingExtension && SnykSolutionService.Instance.IsSolutionOpen())
{
    haveBegunLoadingExtension = true;
    ...
```

Snyk.VisualStudio.Extension.2022/SnykVSPackage.cs
```diff
@@ -64,6 +65,7 @@ public sealed class SnykVSPackage : AsyncPackage, ISnykOptionsProvider
             new TaskCompletionSource<bool>();
 
         public static SnykVSPackage Instance;
+        private readonly SemaphoreSlim semaphore = new SemaphoreSlim(1, 1);
```
andrewrobinsonhodges-snyk, line 68: nitpick - Might be better to name this languageClientInitSemapore or similar, to avoid potential confusion if further semaphores need to be added to the class in the future.

Snyk.VisualStudio.Extension.2022/SnykVSPackage.cs
ShawkyZ: I would rather wait maybe in case isSolutionOrFolderOpen is true and no file tabs are loaded yet as an extra check before we attempt to open our dummy .cs file to enforce languageClient initialization

andrewrobinsonhodges-snyk (approved):
We discussed via Tuple - this is likely the best we can do given the limitations of the VS callbacks.

There is likely some future work to try and replace the delay, but that's lower priority than getting the init working, which this does.

Approved.

PR #478, opened 2026-05-14
feat: migrate to webview2 for all html rendering [IDE-1707]

andrewrobinsonhodges-snyk:
### Description

This PR replaces the legacy WPF WebBrowser control (IE/MSHTML) with Microsoft.Web.WebView2 (Chromium) across all of our  HTML-hosting surfaces (settings dialog, issue description panel, and scan summary panel), fixing the long-standing DPI scaling bug.

The IE ObjectForScripting bridge is replaced with a single JS↔C#  routing path: window.X(...) → chrome.webview.postMessage → WebView2MessageDispatcher → typed handler, so the LS HTML's existing window-level bridge surface works against WebView2.
  
The tool window's description + summary panels share an environment, and settings runs its own (the settings DialogWindow uses a different mechanism to instantiate the webview instance which makes it incompatible. This is a well known limitation. Environments acre created per-process ID (to allow for multiple instances of VS to run simultaneously), and old environments (e.g., from crashed instances) are cleaned up automatically) 

### Checklist

- [x] Read and understood the [Code of Conduct](https://github.com/snyk/snyk-visual-studio-plugin/blob/main/CODE_OF_CONDUCT.md) and [Contributing Guidelines](https://github.com/snyk/snyk-visual-studio-plugin/blob/main/CONTRIBUTING.md).
- [x] Linted
- [ ] CHANGELOG.md updated

### Screenshots / GIFs


Snyk.VisualStudio.Extension.2022/Resources/ScanSummaryInit.html
```diff
@@ -2,39 +2,33 @@
 <html>
 <head>
   <meta http-equiv='Content-Type' content='text/html; charset=unicode' />
```
bastiandoetsch, line 4: Verification Skill (AI): [Critical] The `static-resource-checks` CI job has been failing on every commit to this PR because `ScanSummaryInit.html` no longer matches the reference copy in `snyk/snyk-ls`. The CI check computes SHA-512 of this file and compares it to `https://raw.githubusercontent.com/snyk/snyk-ls/refs/heads/main/shared_ide_resources/ui/html/ScanSummaryInit.html`. The local copy was intentionally modified (removed Google Fonts, IE-compat meta, resized CSS), but the reference in snyk-ls was not updated in lock-step. Either (a) the changes need to be merged into snyk-ls first and this file synced from there, or (b) the modifications should not be made here at all. Test to write first: a test that hashes the local file and asserts it equals the hash of the upstream reference.

Snyk.VisualStudio.Extension.2022/UI/Html/WebView2Host.cs
```diff
@@ -0,0 +1,265 @@
...
+            {
+                // Process has exited between GetProcessById and the HasExited check.
+                return false;
+            }
+        }
+
+        public async Task InitializeAsync()
+        {
+            if (_initialized) return;
+            _initialized = true;
```
bastiandoetsch, line 212: Verification Skill (AI): [Should Fix] `_initialized = true` is set before the async work begins. If `EnsureCoreWebView2Async`, `AddScriptToExecuteOnDocumentCreatedAsync`, or any other awaited call throws, `_initialized` is already `true`. A subsequent call to `InitializeAsync()` returns immediately (line 191 guard), leaving `_readyTcs` faulted and every future `NavigateAsync`/`ExecuteScriptAsync` awaiting a permanently faulted task. The flag should only be set to `true` after the try block succeeds, or the class should allow a retry by clearing `_initialized` in the catch and re-creating `_readyTcs`. Test to write first: call `InitializeAsync()` on a host whose environment creation is mocked to fail; then call `InitializeAsync()` again; assert the second call attempts initialization rather than silently returning.

Snyk.VisualStudio.Extension.2022/UI/Html/WebView2Host.cs
```diff
@@ -0,0 +1,265 @@
...
+        {
+            lock (EnvironmentGate)
+            {
+                if (!Environments.TryGetValue(userDataFolder, out var task))
+                {
+                    task = CoreWebView2Environment.CreateAsync(
+                        browserExecutableFolder: null,
+                        userDataFolder: userDataFolder,
+                        options: null);
+                    Environments[userDataFolder] = task;
```
bastiandoetsch: Verification Skill (AI): [Should Fix] `GetOrCreateEnvironmentAsync` caches the `Task<CoreWebView2Environment>` returned by `CoreWebView2Environment.CreateAsync` before it has completed. If that task faults (e.g. WebView2 runtime not installed, user-data folder locked), the faulted task is stored in `Environments` permanently. Every subsequent call for the same folder returns the same faulted task, making the failure unrecoverable for the lifetime of the VS process. The cache should only store the task after it has completed successfully, or failed tasks should be evicted on next access. Test to write first: mock `CreateAsync` to fault; call `GetOrCreateEnvironmentAsync` twice; assert the second call retries rather than returning the original faulted task.

Snyk.VisualStudio.Extension.2022/UI/Toolwindow/HtmlDescriptionPanel.xaml.cs
```diff
@@ -1,65 +1,125 @@
...
+                {
+                    Logger.Error(ex, "WebView2 initialization failed in description panel");
+                }
+            }).FireAndForget();
         }
 
-        private void HtmlViewerOnLoadCompleted(object sender, NavigationEventArgs e)
+        // The provider's init script (link-click interceptors, etc.) has to run after each
+        // navigation completes, since NavigateAsync replaces the document.
+        private async void OnNavigationCompleted(object sender, CoreWebView2NavigationCompletedEventArgs e)
```
bastiandoetsch, line 73: Verification Skill (AI): [Should Fix] `OnNavigationCompleted` does not check `e.IsSuccess` before executing `htmlProvider.GetInitScript()`. If WebView2 navigation fails (e.g. the HTML temp file was deleted before WebView2 finished loading, or a CSP violation caused an error page), the init script is executed against an error page document rather than the intended LS HTML. The link-click interceptors and other setup in the init script would be wired to the wrong page. The same issue exists in `SummaryHtmlPanel.OnNavigationCompleted` (line 68). Test to write first: call `OnNavigationCompleted` with `e.IsSuccess == false`; assert `ExecuteScriptAsync` is never called.

Snyk.VisualStudio.Extension.2022/UI/Html/HtmlSettingsScriptingBridge.cs
```diff
@@ -78,15 +83,13 @@ public void __saveIdeConfig__(string jsonString)
                     // Persist all settings to storage at the end
                     // This triggers SettingsChanged event which notifies Language Server
                     OptionsManager.Save(Options, triggerSettingsChangedEvent: true);
-
-                    IsSaveComplete = true;
                 });
+                saveCompletionTcs.TrySetResult(true);
```
bastiandoetsch: Verification Skill (AI): [Should Fix] `saveCompletionTcs.TrySetResult(true)` is called outside the `ThreadHelper.JoinableTaskFactory.Run(async () => { ... })` block (line 87), while the actual persistence work happens synchronously inside that block. This is structurally correct when `Run()` does not throw — but if `ParseAndSaveConfigAsync` throws inside the `Run` block, the exception propagates to the `catch`, which calls `TrySetResult(false)`. The save-result signal therefore reports `false` (failure), which is correct. However, **if the `Run` block succeeds but `OptionsManager.Save()` on line 85 throws**, the exception propagates out of `Run()`, and `saveCompletionTcs.TrySetResult(true)` on line 87 is never reached — instead the `catch` at line 89 fires with `TrySetResult(false)`. This is the correct behaviour, so the logic is sound. Minor clarity issue: the comment on line 84 (`// Persist all settings...`) sits inside the lambda but the TCS completion is outside, which is misleading. Consider moving `TrySetResult(true)` to the end of the `Run` block for clarity.

Snyk.VisualStudio.Extension.2022/Settings/DebugHtmlSettingsWindow.xaml.cs
```diff
@@ -4,32 +4,39 @@
...
+        /// Expose Chromium DevTools (F12) and the default right-click context menu so
+        /// developers can inspect JS errors and the DOM while iterating on the LS HTML.
+        /// </summary>
+        protected override bool DeveloperToolsEnabled => true;
+
         /// <summary>
         /// Path to the local HTML file to load for testing.
         /// Generate this file by running in snyk-ls: go run scripts/config-dialog/main.go > config_output.html
         /// </summary>
         private const string DEBUG_HTML_FILE_PATH = @"C:\Mac\Home\Documents\Snyk\snyk-ls\config_output.html";
```
bastiandoetsch, line 26: Verification Skill (AI): [Suggestion] `DEBUG_HTML_FILE_PATH` is hardcoded to `C:\Mac\Home\Documents\Snyk\snyk-ls\config_output.html` — a developer-machine-specific path. This will be a dead reference for any other developer who tries to use `DebugHtmlSettingsWindow`. Consider using an environment variable (e.g. `Environment.GetEnvironmentVariable("SNYK_DEBUG_HTML_PATH")`) with a clear fallback error message, which makes the class usable across the team without a code change.

bastiandoetsch:
Verification Skill (AI): [Should Fix] **HtmlDescriptionPanel and SummaryHtmlPanel do not dispose their WebView2Host.** Both panels hold a `WebView2Host host` that wraps a `WebView2NavigationPreparer` (which owns temp files on disk) and subscribes to `CoreWebView2.WebMessageReceived`. Neither `HtmlDescriptionPanel` nor `SummaryHtmlPanel` implements `IDisposable` or subscribes to `Unloaded` to call `host.Dispose()`. This leaves temp HTML files in `%LOCALAPPDATA%\Snyk\WebView2\<pid>\toolwindow\scratch` undeleted when the tool window is closed, and the event subscription leaks until the VS process exits. `HtmlSettingsWindow` correctly disposes the host in `OnClosed` — the same pattern should be applied to both panels. Test to write first: create the panel, navigate twice with oversized HTML (to generate temp files), close the tool window, assert the scratch directory is empty.

Snyk.VisualStudio.Extension.2022/Settings/DebugHtmlSettingsWindow.xaml.cs
andrewrobinsonhodges-snyk: This exists on main - not part of this change. Worth fixing, but I'd like to keep this PR tightly scoped.

Snyk.VisualStudio.Extension.2022/UI/Html/WebView2Host.cs
andrewrobinsonhodges-snyk: Fixed - Moved inside the async block with a guard against multiple access.

Snyk.VisualStudio.Extension.2022/UI/Toolwindow/HtmlDescriptionPanel.xaml.cs
andrewrobinsonhodges-snyk: Fixed - Checked at line 77 below.

Snyk.VisualStudio.Extension.2022/Settings/DebugHtmlSettingsWindow.xaml.cs
```diff
@@ -4,32 +4,39 @@
...
+    /// bindings at page-load time, so the real scripting bridge — though wired — never
+    /// receives messages.
     /// </summary>
     public class DebugHtmlSettingsWindow : HtmlSettingsWindow
     {
+        /// <summary>
+        /// Expose Chromium DevTools (F12) and the default right-click context menu so
+        /// developers can inspect JS errors and the DOM while iterating on the LS HTML.
+        /// </summary>
+        protected override bool DeveloperToolsEnabled => true;
```
bastiandoetsch: Nitpick: probably should be deactivated for release

Snyk.VisualStudio.Extension.2022/Settings/DebugHtmlSettingsWindow.xaml.cs
```diff
@@ -57,24 +64,21 @@ public static async Task OpenDebugWindowIfEnabledAsync(ISnykServiceProvider serv
...
+
         /// <summary>
         /// Initializes a new debug HTML settings window.
         /// Passes serviceProvider to base HtmlSettingsWindow constructor.
         /// </summary>
         public DebugHtmlSettingsWindow(ISnykServiceProvider serviceProvider)
             : base(serviceProvider)
         {
-            // Enable script error visibility for debugging
-            wbHandler.ScriptErrorsSuppressed = false;
```
bastiandoetsch, line 69: Shouldn't this be tied to debug logging enabled, instead of being removed completely?

Snyk.VisualStudio.Extension.2022/UI/Html/BaseHtmlProvider.cs
```diff
@@ -27,7 +27,7 @@ public virtual string GetInitScript()
                     var links = document.querySelectorAll('a');
                     for(var i = 0; i < links.length; i++) {
                         links[i].onclick = function() {
-                            window.external.OpenLink(this.href);
+                            window.OpenLink(this.href);
```
bastiandoetsch, line 33: Why this? Is this just a package change or does the behavior change from opening in an external browser to an internal?

Snyk.VisualStudio.Extension.2022/UI/Html/ExecuteCommandBridge.cs
```diff
@@ -68,11 +70,47 @@ public static string BuildClientScript()
...
             ";
         }
 
+        /// <summary>
+        /// Builds the JS that invokes <c>window.setAuthToken(token, apiUrl)</c> on the page,
+        /// guarded by a <c>typeof</c> check so a missing function logs rather than throws.
+        /// Used by the settings panel when the Language Server pushes an updated auth token
+        /// (e.g. after an OAuth round-trip).
+        /// </summary>
+        public static string BuildSetAuthTokenScript(string token, string apiUrl)
```
bastiandoetsch, line 84: Why do we need this? Isn't that already provided by the config dialog? If not, why? If it's pre-existing, we should file a follow-up cleanup task.

Snyk.VisualStudio.Extension.2022/UI/Html/ExecuteCommandBridge.cs
```diff
@@ -68,11 +70,47 @@ public static string BuildClientScript()
...
+            ";
+        }
+
+        /// <summary>
+        /// Builds the JS that pops a pending callback from <c>window.__ideCallbacks__</c> and
+        /// invokes it with the LS command's result. The <paramref name="callbackId"/> must have
+        /// already been validated with <see cref="IsValidCallbackId"/>; we still escape it as a
+        /// belt-and-braces guard against accidental injection.
+        /// </summary>
+        public static string BuildCommandCallbackScript(string callbackId, string resultJson)
```
bastiandoetsch, line 105: Same here

Snyk.VisualStudio.Extension.2022/UI/Html/WebView2BridgeBindings.cs
```diff
@@ -0,0 +1,74 @@
...
+            public string Name { get; }
+            public IReadOnlyList<string> Parameters { get; }
+        }
+
+        // window.__ideExecuteCommand__ is intentionally NOT defined here. The LS HTML
+        // calls it with a JS callback function as the third argument, which cannot survive
+        // a raw postMessage round-trip. The settings panel registers
+        // ExecuteCommandBridge.BuildClientScript() as an additional init script, which
+        // defines a proper wrapper that owns a client-side callback-id map.
+        private static readonly IReadOnlyList<Method> Definitions = new[]
```
bastiandoetsch, line 35: Interesting. Is this a general problem and should we fix this centrally?

Snyk.VisualStudio.Extension.2022/UI/Toolwindow/SnykToolWindowControl.xaml.cs
```diff
@@ -857,19 +859,15 @@ private void HandleRootTreeNodeSelected()
 <html>
 <head>
   <meta http-equiv='Content-Type' content='text/html; charset=unicode' />
-  <meta http-equiv='X-UA-Compatible' content='IE=edge' />
   <meta charset=""utf-8"" />
   <meta name=""viewport"" content=""width=device-width, initial-scale=1.0"" />
   <meta http-equiv=""Content-Security-Policy""
-        content=""style-src 'self' 'nonce-${{nonce}}' 'nonce-ideNonce' https://fonts.googleapis.com;
-        script-src 'nonce-${{nonce}}' https://fonts.googleapis.com;
```
bastiandoetsch, line 865: If we need this file at all (why?) - Why are fonts removed? I think we need them.

bastiandoetsch (approved):

Snyk.VisualStudio.Extension.2022/UI/Html/ExecuteCommandBridge.cs
andrewrobinsonhodges-snyk: As above.

Snyk.VisualStudio.Extension.2022/UI/Toolwindow/SnykToolWindowControl.xaml.cs
andrewrobinsonhodges-snyk: They were never actually used; we fetched them, but then subbed in the Windows fonts directly. I jsut removed the references here to avoid fetching them.

PR #493, opened 2026-06-01
feat: enable html settings by default [IDE-1651]

andrewrobinsonhodges-snyk:
### Description

Switched to the wEbView2 Edge renderer for all HTML rendering.
Enabled HTML settings page by default

Note the HTML rendering uses two edge processes; one shared by all parts of the Snyk panel - summary panel, tree view (once added) and info view, and one for the settings page. This is because the dialog containing the settings uses a different native UI rendering framework, so cannot reuse the cache from the other process.

Added handlers for snake case settings vars

I moved Auth handling to a separate service, as it was an integral part of the old settings page and now needs to be separate.

### Checklist

- [ ] Linted
- [ ] CHANGELOG.md updated

### Screenshots / GIFs


Snyk.VisualStudio.Extension.2022/Authentication/AuthenticationFlowService.cs
```diff
@@ -0,0 +1,113 @@
...
+                }
+
+                Logger.Information("Api token is invalid. Attempting to authenticate via snyk auth");
+
+                ThreadHelper.JoinableTaskFactory.RunAsync(async () =>
+                {
+                    await serviceProvider.LanguageClientManager.InvokeLogout(SnykVSPackage.Instance.DisposalToken);
+                    Logger.Information("Invoking InvokeLogin for auth");
+                    await serviceProvider.LanguageClientManager.InvokeLogin(SnykVSPackage.Instance.DisposalToken);
+                }).FireAndForget();
```
nick-y-snyk: **Suggestion: fire-and-forget InvokeLogout/InvokeLogin has no inner error handling**

The outer `catch (Exception e)` below wraps the synchronous setup code but does **not** cover this async lambda — `RunAsync(...).FireAndForget()` dispatches the work and returns immediately. Any LS exception thrown inside (JSON-RPC error, `ObjectDisposedException` if the client tears down, cancellation) is unobserved and unlogged. The comment on the outer catch ("Couldn't execute Invoke Login through LS") is therefore misleading: it will never fire for actual LS failures.

```csharp
ThreadHelper.JoinableTaskFactory.RunAsync(async () =>
{
    try
    {
        await serviceProvider.LanguageClientManager.InvokeLogout(SnykVSPackage.Instance.DisposalToken);
        Logger.Information("Invoking InvokeLogin for auth");
        await serviceProvider.LanguageClientManager.InvokeLogin(SnykVSPackage.Instance.DisposalToken);
    }
    catch (Exception ex)
    {
        Logger.Error(ex, "Failed to invoke Logout/Login through Language Server");
    }
}).FireAndForget();
```

Snyk.VisualStudio.Extension.2022/Service/SnykService.cs
```diff
@@ -36,6 +37,8 @@ public class SnykService : ISnykServiceProvider, ISnykService
 
         private SnykOptionsManager snykOptionsManager;
         private SnykFeatureFlagService featureFlagService;
+        private IAuthenticationFlowService authenticationFlowService;
```
nick-y-snyk: **Suggestion: DCL field should be `volatile`**

The property getter uses double-checked locking — first read of `authenticationFlowService` at line ~70 is outside the `lock`, and the comment explicitly documents the intent to guard concurrent UI-thread + LS-background-thread access. Without `volatile`, the lock-free first read has no memory barrier and can legally observe a non-null but not-yet-fully-published reference under the ECMA CLI memory model.

One-liner fix:
```csharp
private volatile IAuthenticationFlowService authenticationFlowService;
```

Or replace DCL with `Lazy<T>`:
```csharp
private readonly Lazy<IAuthenticationFlowService> authenticationFlowService =
    new Lazy<IAuthenticationFlowService>(
        () => new AuthenticationFlowService(this),
        LazyThreadSafetyMode.ExecutionAndPublication);

public IAuthenticationFlowService AuthenticationFlowService => authenticationFlowService.Value;
```

nick-y-snyk (commented):
**Important: null config in `ParseAndSaveConfigAsync` silently reports save success** (`HtmlSettingsScriptingBridge.cs` line 196)

`JsonConvert.DeserializeObject<IdeConfigData>` returns `null` when the JSON maps to nothing. When that happens `ParseAndSaveConfigAsync` returns early — no settings applied — but the outer `RunAsync` block continues to `OptionsManager.Save` and signals `saveCompletionTcs.TrySetResult(true)`. User clicks Apply, sees no error, all changes silently discarded.

Throw instead so the outer `catch` signals `TrySetResult(false)` and the user sees the failure dialog:
```csharp
var config = JsonConvert.DeserializeObject<IdeConfigData>(jsonString);
if (config == null)
{
    Logger.Error("Settings payload deserialized to null; raw={Preview}",
        jsonString?.Substring(0, Math.Min(200, jsonString?.Length ?? 0)));
    throw new InvalidOperationException("Settings payload deserialized to null");
}
```

Snyk.VisualStudio.Extension.2022/Language/SnykLanguageClientCustomTarget.cs
```diff
@@ -218,34 +218,9 @@ public async Task OnFolderConfig(JToken arg)
...
-                {
-                    return folderConfig;
-                }
-            }
-
-            return null;
-        }
+        // Use FolderConfigMatcher to find the first matching folder config for the current solution path.
+        private FolderConfig FindMatchingFolderConfig(List<FolderConfig> folderConfigs, string currentSolutionPath) =>
+            FolderConfigMatcher.FindFirstMatching(folderConfigs, currentSolutionPath);
```
nick-y-snyk: **TODO: remove `OnFolderConfig` / `$/snyk.folderConfigs` handling entirely**

IntelliJ and VSCode do not handle `$/snyk.folderConfigs` as a separate notification at all — both receive folder configs exclusively inside `$/snyk.configuration` (the `folderConfigs` array in that payload). This handler appears to be a VS-only legacy path that predates the `$/snyk.configuration` integration.

Additionally, the `GetSolutionFolderAsync` + `FolderConfigMatcher` logic used here to find the "matching" folder config is probably wrong in principle. The LS sends `folderConfigs` keyed by the exact `folderPath` it registered as a workspace folder during initialization — it already knows which path belongs to this workspace. The IDE doesn't need to independently re-derive the solution path and fuzzy-match against the list; it should just use `folderConfigs[0]` directly (VS always has exactly one workspace folder). The fuzzy matching exists to paper over a path normalization mismatch that shouldn't exist if the LS is the source of truth for the paths it sends.

Suggested cleanup:
1. Remove `OnFolderConfig` and `LsConstants.SnykFolderConfig`.
2. Confirm that `$/snyk.configuration` already delivers the same `folderConfigs` payload (check with LS team if `$/snyk.folderConfigs` is still emitted at all).
3. If folder-config application on `$/snyk.configuration` is needed, handle it there directly using `folderConfigs.FirstOrDefault()` — no path matching required.

nick-y-snyk (commented):
See inline comment on `SnykLanguageClientCustomTarget.cs` line 223 for the `OnFolderConfig` / `$/snyk.folderConfigs` removal TODO.

Snyk.VisualStudio.Extension.2022/Language/V25/FolderConfigApplier.cs
```diff
@@ -0,0 +1,103 @@
...
+        {
+            var fc = new FolderConfig { FolderPath = src.FolderPath };
+
+            if (src.Settings == null)
+                return fc;
+
+            foreach (var kvp in src.Settings)
+            {
+                if (kvp.Value?.Value == null)
+                    continue;
```
acke, line 60: 🔴 **Critical**

**[FolderConfigApplier.cs:53-95]** Asymmetric round-trip silently drops per-folder overrides on every `$/snyk.configuration` push. `LsSettingsV25.BuildFolderConfigs` emits 13 per-folder override pflags outbound (`snyk_oss_enabled`, `snyk_code_enabled`, `snyk_iac_enabled`, `snyk_secrets_enabled`, `scan_automatic`, `scan_net_new`, `severity_filter_{critical,high,medium,low}`, `issue_view_{open,ignored}_issues`, `risk_score_threshold`), but `ToFolderConfig`'s switch handles only 8 keys (params/env/orgs/branch/scan-command). When the LS echoes back, the 13 are dropped — the user's per-folder override survives only until the next workspace refresh. Corroborated by 4 lenses (logic, contracts, tests, semantic trace).

**Fix:**
Add `case PflagKeys.SnykOssEnabled` … `case PflagKeys.RiskScoreThreshold` to the switch in `ToFolderConfig`, mirroring `LsSettingsV25.BuildFolderConfigs` exactly. Add a round-trip test in `FolderConfigApplierTests` that runs `BuildFolderConfigs(fc) → ToFolderConfig → FolderConfig` for every override field; let that test drive the missing switch arms.

— AI review comment

Snyk.VisualStudio.Extension.2022/Settings/SnykSettings.cs
```diff
@@ -24,11 +24,6 @@ public SnykSettings()
         /// </summary>
         public string CurrentCliVersion { get; set; }
 
-        /// <summary>
-        /// Gets or sets solution settings dictionary.
-        /// </summary>
-        public IDictionary<int, SnykSolutionSettings> SolutionSettingsDict { get; set; } = new Dictionary<int, SnykSolutionSettings>();
-
         /// <summary>
```
acke, line 28: 🔴 **Critical**

**[SnykSettings.cs:27 + SnykSolutionSettings.cs]** Removal of `SolutionSettingsDict` is a silent-drop migration. Newtonsoft `MissingMemberHandling` is not set to `Error`, so an upgrading user's existing `settings.json` loads cleanly — but every persisted per-solution `AdditionalOptions`, `AdditionalEnv`, `Organization`, `AutoDeterminedOrg`, `PreferredOrg`, `OrgSetByUser`, `IsAllProjectsScanEnabled` is silently discarded on the next `SaveSettingsToFile` (triggered by almost any interaction including OAuth completion). No warning logged. No migration shim. Combined with the `GetLsDebugLevelAsync` regression flagged separately, the first-run scan after upgrade can execute with wrong org / no debug flag / empty CLI args.

**Fix:**
Add a one-time migration in `SnykSettingsLoader.Load` that reads `solutionSettingsDict` from the raw JSON before deserialization, projects the matching solution's values into the new global / per-folder state (or pushes them to LS via `workspace/didChangeConfiguration` once the LS is up), then strips the legacy section on the next save. At minimum, log a warning when a non-empty `solutionSettingsDict` is observed in the loaded JSON so support has visibility.

— AI review comment

Snyk.VisualStudio.Extension.2022/SnykVSPackage.cs
```diff
@@ -50,12 +50,7 @@ namespace Snyk.VisualStudio.Extension
     [ProvideService(typeof(ISnykService), IsAsyncQueryable = true)]
     [ProvideMenuResource("Menus.ctmenu", 1)]
     [ProvideToolWindow(typeof(SnykToolWindow), Style = VsDockStyle.Tabbed)]
-    [ProvideOptionPage(typeof(SnykGeneralOptionsDialogPage), "Snyk", "Account", 1000, 1001, true)]
-    [ProvideOptionPage(typeof(SnykScanOptionsDialogPage), "Snyk", "Scan Configuration", 1000, 1002, true)]
-    [ProvideOptionPage(typeof(SnykSolutionOptionsDialogPage), "Snyk", "Solution settings", 1000, 1003, true)]
-    [ProvideOptionPage(typeof(SnykCliOptionsDialogPage), "Snyk", "CLI settings", 1000, 1004, true)]
-    [ProvideOptionPage(typeof(SnykExperimentalDialogPage), "Snyk", "Experimental", 1000, 1005, true)]
-    [ProvideOptionPage(typeof(SnykUserExperienceDialogPage), "Snyk", "User Experience", 1000, 1006, true)]
+    [ProvideOptionPage(typeof(HtmlSettingsDialogPage), "Snyk", "General", 1000, 1001, true)]
```
acke, line 55: 🟠 **High**

**[SnykVSPackage.cs:53 + Settings/HtmlSettings*.cs]** HTML settings ON by default with no rollback path. All six legacy `[ProvideOptionPage]` registrations and their DialogPage / UserControl / .resx files are physically deleted. `settings-fallback.html` still requires WebView2 to render. If WebView2 init fails (missing Edge WebView2 Runtime, corporate-locked Edge, sandbox issues, HRESULT 0x8007139F that `WebView2Host` itself documents), the user has zero way to change Snyk settings — auth, endpoint, CLI path, organization — all unreachable. This is the first release that flips the flag; it's also the release that deletes the escape hatch.

**Fix:**
Ship at least one of: (a) keep `SnykGeneralOptionsDialogPage` for one more release behind a registry / `App.config` flag as a hidden fallback; (b) add a `Tools → Snyk → Open Settings JSON` command that opens the on-disk `settings.json` in VS's editor as a manual escape hatch; (c) declare the Edge WebView2 Runtime as a hard VSIX prerequisite and verify at package load with a clear actionable error.

— AI review comment

Snyk.VisualStudio.Extension.2022/AuthDialogWindow.xaml.cs
```diff
@@ -10,12 +10,45 @@ public partial class AuthDialogWindow : DialogWindow
     {
         public static AuthDialogWindow Instance { get; } = new AuthDialogWindow();
 
+        // Guards the show/hide race in the auth flow: login runs fire-and-forget while the modal
+        // ShowDialog() is started separately, so a fast success/failure can call Hide() *before*
```
acke: 🟠 **High**

**[AuthDialogWindow.xaml.cs:11-55 + Authentication/AuthenticationFlowService.cs:60-99]** New auth race-guard documents an invariant it doesn't satisfy, and the mechanism may not actually unblock the modal pump. Four independent issues:

1. The class comment says "all callers SwitchToMainThread first" — but `AuthenticationFlowService.Authenticate()` calls `AuthDialogWindow.Instance.ArmForShow()` synchronously, and the primary caller chain `MessagePanel.RunTestCodeNowAsync → FireAndForget → await GetSolutionFolderAsync() → Authenticate()` lands the continuation on the thread pool. `ArmForShow` definitely runs off-UI.
2. `HideForAuthResult()` and the `Closing` handler set `Visibility = Visibility.Hidden`. `Window.ShowDialog()` on `Microsoft.VisualStudio.PlatformUI.DialogWindow` only returns when `DialogResult` is set or `Close()` is called — assigning `Hidden` does NOT exit the modal pump. The "stuck dialog" bug this guard exists to fix can recur.
3. TOCTOU in `ShowDialogForAuth`: between the `if (authResultArrived) return;` check and `ShowDialog()`, a fast LS callback hitting `HideForAuthResult` (then setting `authResultArrived = true`) doesn't help — once `ShowDialog()` has entered its pump, only `Close()` exits.
4. `Authenticate()` has no re-entrancy guard; a second tool-window button click while a flow is in progress hits `Window.ShowDialog()` again, which throws `InvalidOperationException("The window is already visible")`.

Corroborated by 4 lenses (logic, ai-patterns, adversarial, tests).

**Fix:**
- Wrap `ArmForShow` in `JTF.Run(async () => { await SwitchToMainThreadAsync(); ArmForShow(); })` (or push the switch inside `ArmForShow`) so the documented invariant holds.
- Replace both `this.Visibility = Visibility.Hidden` writes with `this.Hide()` (or expose a real `Close()` if you need the dialog to dispose) so the modal pump actually unblocks.
- In `ShowDialogForAuth`, also short-circuit on `IsVisible == true` to defend against re-entrant `Authenticate()` calls.
- Drop the `volatile` keyword once the UI-thread invariant is real, and add `Assumes.True(ThreadHelper.CheckAccess())` in each `AuthDialogWindow` method.

— AI review comment

Snyk.VisualStudio.Extension.2022/UI/Html/HtmlSettingsScriptingBridge.cs
```diff
@@ -165,13 +173,46 @@ await ExecuteCommandBridge.DispatchAsync(
         public void __ideSaveAttemptFinished__(string status)
```
acke, line 214: 🟠 **High**

**[WebView2Host.cs:235 + HtmlSettingsScriptingBridge.cs:142-148]** (Anchored at HtmlSettingsScriptingBridge.cs:173 because the actual lines are outside this PR's diff hunks — see actual line numbers in the body.) Two compounding WebView2 security gaps:

1. **No off-origin navigation gating.** Neither WebView2 host subscribes to `CoreWebView2.NavigationStarting` / `NewWindowRequested` / `AddWebResourceRequestedFilter`. LS-served HTML (or any page that ever reaches the control) can navigate to arbitrary off-origin URLs while the C# bridges (`__saveIdeConfig__`, `__ideExecuteCommand__`, `OpenLink`) stay wired and callable.
2. **`snyk.login` bridge writes `IgnoreUnknownCA` + `CustomEndpoint` from JS-supplied args with zero validation.** A compromised or off-origin page can disable TLS verification and repoint the API host before the next token round-trip.

**Fix:**
- In `WebView2Host`, subscribe to `CoreWebView2.NavigationStarting` and `NewWindowRequested`; cancel any navigation outside an allowlist (about:blank + the loaded temp-file URI).
- In `HtmlSettingsScriptingBridge.__ideExecuteCommand__`'s `snyk.login` branch (line ~148), validate `CustomEndpoint` is `https://` and host is in a configured allowlist (or at least `snyk.io` / corporate proxy host); refuse `IgnoreUnknownCA = true` unless gated by an explicit IDE-side confirmation.

— AI review comment

Snyk.VisualStudio.Extension.2022/SnykVSPackage.cs
```diff
@@ -225,6 +213,13 @@ protected override async Task InitializeAsync(CancellationToken cancellationToke
         /// <param name="disposing">Bool.</param>
```
acke, line 236: 🟠 **High**

**[SnykVSPackage.cs:53]** Six `[ProvideOptionPage]` registrations deleted with no profile-cleanup migration. The prior IDs are registered in users' VS profiles (`HKCU\Software\Microsoft\VisualStudio\<ver>\DialogPage\Snyk.VisualStudio.Extension.Settings.Snyk*DialogPage`). On first launch of the new VSIX, the Tools → Options "Snyk" node will show the new `General` page **plus** stale VS-cached entries for the now-missing `SnykScanOptionsDialogPage`, `SnykCliOptionsDialogPage`, `SnykExperimentalDialogPage`, `SnykSolutionOptionsDialogPage`, `SnykUserExperienceDialogPage`. Clicking any of those swallows a `TypeLoadException`.

**Fix:**
Add a one-shot migration in `SnykVSPackage.InitializeAsync` that clears the stale `HKCU\Software\Microsoft\VisualStudio\<ver>\DialogPage\Snyk.VisualStudio.Extension.Settings.Snyk*DialogPage` keys on first run of the new version, or document a clean-install requirement explicitly in release notes.

— AI review comment

Snyk.VisualStudio.Extension.2022/Language/LsConstants.cs
```diff
@@ -11,7 +11,6 @@ public static class LsConstants
         public const string SnykCliPath = "$/snyk.isAvailableCli";
         public const string SnykAddTrustedFolders = "$/snyk.addTrustedFolders";
         public const string SnykScan = "$/snyk.scan";
-        public const string SnykFolderConfig = "$/snyk.folderConfigs";
         // This notification is needed because we are sending Issue data in the Diagnostic.Data field and Visual Studio filters it out.
```
acke, line 14: 🟠 **High**

**[LsConstants.cs:14 + Language/SnykLanguageClientCustomTarget.cs]** `$/snyk.folderConfigs` constant and its `[JsonRpcMethod]` handler are both removed. snyk-ls builds older than the version that switched to surfacing folder configs only via `$/snyk.configuration` will still send `$/snyk.folderConfigs` notifications, which now hit JsonRpc's no-handler path and surface as `MethodNotFound` log noise (or RPC error responses if the LS sends as request). Compatibility window shrinks unilaterally.

**Fix:**
Keep the constant and add a no-op handler (`return Task.CompletedTask;` with a one-line "deprecated; LS now sends via $/snyk.configuration" log) until the minimum-supported snyk-ls version is known to send folder configs only via `$/snyk.configuration`.

— AI review comment

Snyk.VisualStudio.Extension.2022/Settings/HtmlSettingsDialogPage.cs
```diff
@@ -0,0 +1,240 @@
...
+
+        protected override void OnApply(PageApplyEventArgs e)
+        {
+            // Re-entrancy guard. Dispatcher.PushFrame below pumps the *full* dispatcher, so the
+            // modal Options dialog can deliver a second Apply/OK click while the first save is
+            // still in flight. Letting that through would re-enter SaveAsync → BeginSave and reset
+            // the bridge's save-completion source mid-flight, stranding the first save (it awaits
+            // the now-orphaned TCS until its 5s timeout) and racing OptionsManager.Save writes.
+            // OnApply only runs on the UI thread, and the pump is the only re-entry path, so a
+            // plain bool is sufficient — no locking needed. Drop the duplicate and keep the page
```
acke, line 139: 🟠 **High**

**[HtmlSettingsDialogPage.cs:60-65, 139-208]** `OnApply` save path can silently lose the user's click. `Control` getter calls `EnsureFreshControl()` which returns without assigning `control` when `serviceProvider == null` (Initialize hasn't run yet); `OnApply` then dereferences `activeControl.SaveAsync()` and NREs. The outer try/catch sets `ApplyKind.CancelNoNavigate` correctly so the user stays on the page, but the silent failure is user-hostile. Worse: `OnApply` reads `var activeControl = Control;` (which can re-create the control) **before** setting `saveInProgress = true` — a concurrent `HostBorder.IsVisibleChanged` on the same dispatcher tick can swap the control out from under the resolved `activeControl`, and `Dispatcher.PushFrame` pumping the full dispatcher just below makes the window real.

**Fix:**
- Add `if (Control == null) { e.ApplyBehavior = ApplyKind.CancelNoNavigate; logger.Warning(...); return; }` at the top of `OnApply` so the user sees a clear failure path instead of a silent swallow.
- Set `saveInProgress = true` BEFORE the `var activeControl = Control;` read; if `Control` returns null in that ordering, fail the save fast.

— AI review comment

Snyk.VisualStudio.Extension.2022/Language/SnykLanguageClient.cs
```diff
@@ -218,16 +219,21 @@ public Task OnServerInitializedAsync()
             return Task.CompletedTask;
         }
 
-        private async Task<string> GetLsDebugLevelAsync(ISnykOptionsManager optionsManger)
+        private async Task<string> GetLsDebugLevelAsync()
```
acke: 🟡 **Medium**

**[SnykLanguageClient.cs:222-238]** `GetLsDebugLevelAsync` is called at LS spawn from `Options.FolderConfigs`, which is empty before the first `$/snyk.configuration` arrives. Users who set `--debug` / `-d` in additional CLI parameters now get info-level logs every session. Pre-PR implementation read from `optionsManager.GetAdditionalOptionsAsync()` (disk-backed, populated by previous sessions). Same method is also declared `async Task<string>` with no `await` (CS1998).

**Fix:**
Read additional parameters from the raw saved options (`SnykSettingsLoader.Load(...)`) before LS spawn so previously-set debug flags survive across sessions. Drop `async` or add `await Task.Yield()` to silence CS1998.

— AI review comment

Snyk.VisualStudio.Extension.2022/UI/Html/IdeConfigData.cs
```diff
@@ -8,57 +8,79 @@
...
+    /// CLI-only fallback HTML). Both forms emit a flat, snake_case pflag-keyed payload — the
+    /// same key scheme the IDE uses when it sends settings to the Language Server (see
+    /// <see cref="LsSettingsV25"/> / PflagKeys). Every field therefore binds explicitly to its
+    /// snake_case JSON key; relying on default camelCase matching silently drops snake_case keys,
+    /// <para>
+    /// The form sends only the fields that changed, so every property is nullable / reference
+    /// type and the bridge applies it only when present.
+    /// </para>
     /// </summary>
     public class IdeConfigData
```
acke, line 37: 🟡 **Medium**

**[IdeConfigData.cs:21]** JS↔C# wire format flipped from PascalCase + nested `IssueViewOptions` / `FilterSeverity` objects to flat snake_case pflag keys (`snyk_oss_enabled`, `severity_filter_critical`, `issue_view_open_issues`, `api_endpoint`, `scan_automatic`, `scan_net_new`). `IsFallbackForm` is the only field still on the old camelCase. An LS HTML build emitting the old shape (`activateSnykOpenSource`, `filterSeverity`, etc.) deserialises to all-null `IdeConfigData` and silently no-ops every save. Compounds with the false-vs-absent semantics issue flagged in `__saveIdeConfig__`.

**Fix:**
Add a schema-version guard in `__saveIdeConfig__` (reject and log an error if the LS HTML is older than the snake_case revision), or add tolerant fallbacks for the legacy camelCase keys. Document the snake_case contract in `IdeConfigData.cs`.

— AI review comment

Snyk.VisualStudio.Extension.2022/UI/Html/HtmlSettingsScriptingBridge.cs
```diff
@@ -165,13 +173,46 @@ await ExecuteCommandBridge.DispatchAsync(
...
+            // without calling __saveIdeConfig__, so SaveCompletion would otherwise sit unsignalled
+            // until HtmlSettingsControl.SaveAsync's 5s timeout — leaving the Apply button hung and
+            // then surfacing a misleading "check the log" error. A non-"success" status lets
+            // OnApply report the failure immediately and keep the user on the page to fix it.
+            // The success path is still driven by __saveIdeConfig__ → TrySetResult(true);
+            // TrySetResult is first-wins, so a later "success" signal here is a harmless no-op.
+            if (!string.IsNullOrEmpty(status) &&
+                !string.Equals(status, "success", StringComparison.OrdinalIgnoreCase))
+            {
+                saveCompletionTcs.TrySetResult(false);
```
acke: 🟡 **Medium**

**[HtmlSettingsScriptingBridge.cs:200-225 + 402-426 + IdeConfigData RiskScoreThreshold]** `ParseAndSaveConfigAsync` throws on null deserialize (good), but if the LS HTML changes a key name, Newtonsoft yields a non-null `IdeConfigData` with all-null nullable fields. Every `Apply*` helper no-ops, `Save` writes unchanged settings, and `TrySetResult(true)` reports success. User clicks Apply with no effect. Same root issue for severity filters and `RiskScoreThreshold`: implementation requires "user unchecked X" to send `false` and "user did not touch X" to omit — an undocumented and untested contract.

**Fix:**
Track whether any field was applied in `ParseAndSaveConfigAsync`; if zero, log a warning AND fail the save (`TrySetResult(false)`). Document the false-vs-absent contract in `IdeConfigData.cs` and add a captured-payload contract test that pins each severity filter / issue view / risk-score-threshold field.

— AI review comment

Snyk.VisualStudio.Extension.2022/UI/Html/HtmlSettingsScriptingBridge.cs
```diff
@@ -91,13 +99,13 @@ public void __saveIdeConfig__(string jsonString)
                     OptionsManager.Save(Options, triggerSettingsChangedEvent: true);
 
                     saveCompletionTcs.TrySetResult(true);
```
acke: 🟡 **Medium**

**[HtmlSettingsScriptingBridge.cs:84-110]** Save-completion plumbing races the LS push. `__saveIdeConfig__` signals `saveCompletionTcs.TrySetResult(true)` AFTER `OptionsManager.Save(triggerSettingsChangedEvent: true)`. The chain `SettingsChanged → OnOptionsSettingsChanged → DidChangeConfigurationAsync` is itself FireAndForget — if `DidChangeConfigurationAsync` faults AFTER `OnApply` closes the dialog, the user sees "saved" but the LS never gets the new settings. Separately: `saveCompletionTcs` is a mutable field captured by closure; if `BeginSave()` is invoked before the in-flight `TrySetResult`, the lambda signals a freshly-reset TCS instead of the one `SaveAsync` is awaiting.

**Fix:**
- Snapshot the TCS into a local at lambda entry (`var tcs = saveCompletionTcs;`) so each save attempt resolves its own TCS.
- Either await the LS push inside `OnApply` (with timeout / error reporting), or surface failures via the Snyk error notification panel.

— AI review comment

Snyk.VisualStudio.Extension.2022/UI/Html/HtmlSettingsScriptingBridge.cs
```diff
@@ -165,13 +173,46 @@ await ExecuteCommandBridge.DispatchAsync(
         public void __ideSaveAttemptFinished__(string status)
         {
             Logger.Information("Save attempt finished with status: {Status}", status);
+
+            // Fast-fail the save when the page reports it did NOT persist (e.g. client-side
+            // validation rejected the CLI version). In that case getAndSaveIdeConfig() returns
```
acke, line 225: 🟡 **Medium**

**[HtmlSettingsScriptingBridge.cs:144-148]** (Anchored on a nearby in-diff line — actual issue at line 148.) `__ideExecuteCommand__` reads `args[2] is bool b ? b : Convert.ToBoolean(args[2])`; `Convert.ToBoolean(null)` throws `InvalidCastException` and `Convert.ToBoolean("yes")` throws `FormatException`. The outer try / catch logs a Warning and proceeds — user sees a silently-not-applied SSL toggle. Also: `args[1]?.ToString()` is written verbatim into `Options.CustomEndpoint` with no scheme/host validation (see High security finding for endpoint validation).

**Fix:**
Replace with explicit handling: accept `bool` / `"true"` / `"false"` / `1` / `0` deterministically; log an Error (not Warning) on unparseable values and propagate failure via `onCommandResult`. Validate `CustomEndpoint` per the High security finding.

— AI review comment

Snyk.VisualStudio.Extension.2022/UI/Html/HtmlSettingsScriptingBridge.cs
```diff
@@ -287,9 +329,9 @@ private void ApplyInsecureSetting(IdeConfigData config)
         private void ApplyConnectionSettings(IdeConfigData config)
```
acke, line 492: 🟡 **Medium**

**[HtmlSettingsScriptingBridge.cs:340-344]** (Anchored on a nearby in-diff line — actual issue at lines 340 / 344 in `ApplyConnectionSettings`.) Trimmed-vs-untrimmed token mismatch: `ApplyConnectionSettings` compares `normalizedNewToken \!= normalizedExistingToken` (trimmed) but constructs `new AuthenticationToken(..., config.Token)` from the UN-trimmed value. Leading / trailing whitespace from the form persists into the token store, then downstream `IsValid()` (`Guid.TryParse` / OAuth JSON parse) silently fails.

**Fix:**
Persist `normalizedNewToken` rather than the raw `config.Token`.

— AI review comment

Snyk.VisualStudio.Extension.2022/UI/Html/HtmlSettingsScriptingBridge.cs
```diff
@@ -165,13 +173,46 @@ await ExecuteCommandBridge.DispatchAsync(
...
+                {
+                    loggedPayload = loggedPayload.Substring(0, 200) + "…(truncated)";
+                }
+
+                Logger.Error(
+                    "Settings payload from the settings page deserialized to null; no settings were saved. Payload: {Payload}",
+                    loggedPayload);
+                throw new InvalidOperationException(
+                    "Could not parse settings payload from the settings page; no settings were saved.");
+            }
```
acke, line 262: 🟡 **Medium**

**[HtmlSettingsScriptingBridge.cs:200-225]** `ParseAndSaveConfigAsync` mutates `Options.*` directly through `Apply*` helpers; if any later step throws, the in-memory `Options` retains a half-applied state until restart re-loads from disk — and the disk save is also never written. User ends up with a partially-changed in-memory state that doesn't match the LS or disk.

**Fix:**
Snapshot `Options` (or stage changes in a `SnykSettings` copy / builder) before mutating, and swap into `Options` only after `Save` succeeds; roll back the in-memory state on exception.

— AI review comment

Snyk.VisualStudio.Extension.2022/Language/SnykLanguageClientCustomTarget.cs
```diff
@@ -274,8 +174,9 @@ public async Task OnHasAuthenticated(JToken arg)
```
acke, line 174: 🟡 **Medium**

**[SnykLanguageClientCustomTarget.cs:173 + UI/Toolwindow/SnykScriptManager.cs:43 + UI/Html/ExecuteCommandBridge.cs:37,52 + UI/Html/BaseHtmlProvider.cs:51]** (Anchored here because the related files are pre-existing and not in this PR's diff. They're in scope because the WebView2 migration shipped in this PR widens their attack surface.) Cluster of standalone security mediums:

- `OnHasAuthenticated` writes LS-supplied `apiUrl` straight into `Options.CustomEndpoint` (line 183 in this file) with no scheme / host validation — same exposure as the JS `snyk.login` path.
- `SnykScriptManager.OpenLink` uses `link.StartsWith("http")` (matches `httpx://`, etc.) and hands the raw string to `Process.Start` with default `UseShellExecute=true` — launches arbitrary registered URI handlers.
- `ExecuteCommandBridge.IsAllowedCommand` accepts any `snyk.*` command — relies entirely on the LS for authorization.
- `ExecuteCommandBridge.EscapeForJsString` handles only `\\` and `'`; doesn't escape U+2028 / U+2029, CR / LF, or `</script` — tokens / apiUrl with those characters break out of the single-quoted JS string in `BuildSetAuthTokenScript` / `BuildCommandCallbackScript`.
- `BaseHtmlProvider.GetNonce` builds CSP nonces from `System.Random` (non-CSPRNG) — predictable nonces defeat the page's `style-src 'nonce-...'` policy.

**Fix:**
- Validate `apiUrl` (absolute + https + allowlisted host) before assignment in `OnHasAuthenticated`.
- `Uri.TryCreate(UriKind.Absolute)` + `scheme == "https"` (or http+https) in `OpenLink`; explicit `new ProcessStartInfo(url) { UseShellExecute = true }`.
- Maintain an explicit allowlist of webview-callable LS commands (`snyk.login`, `snyk.logout`, `snyk.workspaceFolders`, …) in `IsAllowedCommand`.
- Replace `EscapeForJsString` with `JsonConvert.SerializeObject(value)` so JS literals are properly quoted including unicode line separators.
- Switch `GetNonce` to `System.Security.Cryptography.RandomNumberGenerator.GetBytes`.

— AI review comment

Snyk.VisualStudio.Extension.2022/Authentication/AuthenticationFlowService.cs
```diff
@@ -0,0 +1,146 @@
+// ABOUTME: Orchestrates IDE-side auth flow (LS readiness check, login/logout, modal auth dialog).
```
acke, line 1: 🟡 **Medium**

**[Authentication/AuthenticationFlowService.cs + AuthDialogWindow.xaml.cs + Settings/HtmlSettingsDialogPage.cs + Settings/HtmlSettingsControl.xaml.cs + UI/Html/WebView2Host.cs]** Zero unit tests for the new race-guard / lifecycle code that exists *because* this PR fixes those exact races. Specifically:

- `AuthenticationFlowService` (146 lines, central orchestrator): no test file at all. Branches for CLI-not-found throw, LS-not-ready early return, AutoScan trigger, PAT skip-modal, LS-call failure → `HandleFailedAuthenticationAsync` are uncovered.
- `AuthDialogWindow.ArmForShow / HideForAuthResult / ShowDialogForAuth` (the literal "fix race condition" commit): no test verifies hide-before-show suppresses show, show-after-arm-only actually shows, or that `ArmForShow` resets a prior result.
- `HtmlSettingsDialogPage.saveInProgress` re-entrancy guard ("OK clicked multiple times" fix): no test for re-entrant call during a pending save, or that `EnsureFreshControl` does not swap controls while a save is in flight.
- `WebView2Host.GetOrCreateEnvironmentAsync` cache + `EvictEnvironmentCache` ("settings page closed and reopened" fix): no test for in-flight task sharing, faulted-entry eviction, or eviction-then-recreate.

These are all the failure modes the PR adds code to prevent — silent regression is the most likely failure.

**Fix:**
Extract the guard logic into testable helpers (e.g. `AuthDialogState`, `WebView2EnvironmentCache`) and add the minimum tests listed above. Each is small and pays for itself the first time the next refactor touches the file.

— AI review comment

acke:
🟡 **Medium**

**[Snyk.VisualStudio.Extension.2022/Language/V25/GlobalSettingsApplier.cs:90-92]** (Posted as PR-level comment because `GlobalSettingsApplier.cs` was added in a prior PR — IDE-1653 — and is not in this PR's diff. The TODO is in scope because it references the same global-settings push path this PR moves under HTML-settings-by-default.) Stale `TODO [IDE-1653]` in production code lists unhandled pflag keys (`SendErrorReports`, `EnableTelemetry`, `SnykCodeApi`, `Path`, `ActivateSnykCode`, `AutomaticAuthentication`). Classic "didn't finish" marker — either these are intentional drops (in which case the TODO is misleading) or they're real bugs that block this PR.

**Fix:**
Triage each listed key: either implement the apply (mirror the matching `BuildGlobalSettings` outbound), or replace the TODO with a comment explaining why each is intentionally not mapped.

— AI review comment

[IDE-1653]: https://snyksec.atlassian.net/browse/IDE-1653?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

Snyk.VisualStudio.Extension.2022/Settings/HtmlSettingsControl.xaml.cs
```diff
@@ -17,39 +17,37 @@
...
+        /// Tracks the Tools→Options <see cref="HtmlSettingsDialogPage"/>'s control while
+        /// the page is in the visual tree.
         /// </summary>
-        protected virtual bool DeveloperToolsEnabled =>
-#if DEBUG
-            true;
-#else
-            false;
-#endif
+        public static HtmlSettingsControl Instance => instance;
```
acke, line 39: 🟡 **Medium**

**[HtmlSettingsControl.xaml.cs:30-39, 79]** `static volatile HtmlSettingsControl Instance` singleton is over-engineered and breaks at-least-once delivery. The only reader is `SnykLanguageClientCustomTarget.OnHasAuthenticated` pushing OAuth tokens into a visible settings page. If `OnHasAuthenticated` fires while `Instance` is null (between `Unloaded` and the next `IsVisibleChanged`), the token push is lost — a guarantee the deleted `HtmlSettingsWindow.Instance` provided. Also has a stale-write window: a slow `IsVisibleChanged` (control A) followed by a fresh `EnsureFreshControl → new HtmlSettingsControl` (control B) lets A's late `IsVisibleChanged` set `instance = A` after B is already in use.

**Fix:**
Route `OnHasAuthenticated` through `SnykVSPackage.Instance.HtmlSettingsDialogPage?.Control` (canonical accessor), queue the latest token in a static field that the next `IsVisibleChanged` consumes, and delete the `instance` static.

— AI review comment

Snyk.VisualStudio.Extension.Tests/Language/SnykLanguageClientCustomTargetTests.cs
```diff
@@ -250,239 +245,228 @@ public async Task OnAddTrustedFolders_ShouldUpdateTrustedFolders()
         }
```
acke, line 246: 🟡 **Medium**

**[Snyk.VisualStudio.Extension.Tests — multiple files]** Cluster of test-quality issues that weaken the safety net:

- `Language/SnykLanguageClientCustomTargetTests.cs:413` — `OnSnykConfiguration_ShouldSaveOptions_WhenValidSettingsReceived` only verifies `Save(...)` was called; never asserts `OssEnabled` was actually applied.
- Same file:371 — `OnSnykConfiguration_ShouldOverwriteByPath_WhenSamePathArrivesTwice` is misnamed; sends path once, not twice.
- Same file:55, 191 — `OnPublishDiagnostics316_ShouldHandleNullUri` / `ShouldHandleEmptyDiagnostics` / `OnHasAuthenticated_WithExistingToken_OnlySavesQuietly` have empty or weak `Assert` blocks.
- `Language/LsSettingsV25Tests.cs:78` — `GetInitializationOptions_OsArchIsGoarchStyle` asserts "not X64-ish" instead of being in the valid set; a future regression returning `unknown` passes.
- `Language/FolderConfigApplierTests.cs:264` — `ToFolderConfig_SkipsBadValueWithoutThrowing` only covers `OrgSetByUser`; `AdditionalParameters` / `LocalBranches` / `ScanCommandConfig` throw paths uncovered.
- `UI/Html/HtmlSettingsScriptingBridgeTest.cs:184` — `IdeSaveAttemptFinished_DoesNotOverrideSuccessfulSave` asserts `SaveCompletion.IsCompleted` without awaiting; flaky to async refactors.
- Same file:179 — no test for `__saveIdeConfig__` failure path (null / whitespace JSON).
- Same file:42 — `IdeExecuteCommand_SnykLogin_*` tests bypass the dispatcher chain entirely; add at least one end-to-end test that constructs `WebView2MessageDispatcher` and dispatches a raw `{method,args}` JSON.
- Same file:451 — `SaveIdeConfig_FolderConfigs_NoExistingConfig_DoesNotThrow` has no assertion.
- `Settings/SnykOptionsManagerAutoOrgTests.cs` reduced to a 3-line stub; the still-shipping `Get/SaveOrganizationAsync` global org helpers now lack any direct test.

**Fix:**
Walk the list; each is a 1-3 line fix. Stricter assertions, awaited completions, and the global-org test are the highest-leverage entries.

— AI review comment

Snyk.VisualStudio.Extension.2022/Language/V25/FolderConfigApplier.cs
andrewrobinsonhodges-snyk: Fixed - all keys now handled and tested.

Snyk.VisualStudio.Extension.2022/SnykVSPackage.cs
andrewrobinsonhodges-snyk: I don't think we want to support both; HTML settings is needed for Config API and we're not adding native UI for this.

Snyk.VisualStudio.Extension.2022/AuthDialogWindow.xaml.cs
andrewrobinsonhodges-snyk: Fixed the re-entrancy condition. I don't think setting visibility to hidden is an issue though - what's the concern here?

Snyk.VisualStudio.Extension.2022/UI/Html/HtmlSettingsScriptingBridge.cs
andrewrobinsonhodges-snyk: Fixed: Added gating

Snyk.VisualStudio.Extension.2022/Language/LsConstants.cs
andrewrobinsonhodges-snyk: We don't support back versioned CLI binaries; I don't think this is worth adding.

Snyk.VisualStudio.Extension.2022/Settings/HtmlSettingsDialogPage.cs
andrewrobinsonhodges-snyk: Fixed.

Snyk.VisualStudio.Extension.2022/Language/SnykLanguageClient.cs
andrewrobinsonhodges-snyk: Fixed

Snyk.VisualStudio.Extension.2022/UI/Html/IdeConfigData.cs
andrewrobinsonhodges-snyk: Added documentation here, but as above, we're not supporting old CLI binaries.

Snyk.VisualStudio.Extension.2022/UI/Html/HtmlSettingsScriptingBridge.cs
andrewrobinsonhodges-snyk: Fixed

Snyk.VisualStudio.Extension.2022/UI/Html/HtmlSettingsScriptingBridge.cs
andrewrobinsonhodges-snyk: Fixed

Snyk.VisualStudio.Extension.2022/Language/SnykLanguageClientCustomTarget.cs
andrewrobinsonhodges-snyk: All fixed as per the suggestions.

Snyk.VisualStudio.Extension.2022/SnykVSPackage.cs
andrewrobinsonhodges-snyk: Surely this is not how the extension works? I don't think anyone should be relying on new VSIX packages deleting keys from what is essentially a different extension. My guess is that the AI has made this up.

Snyk.VisualStudio.Extension.2022/Authentication/AuthenticationFlowService.cs
andrewrobinsonhodges-snyk: Added recommended tests

Snyk.VisualStudio.Extension.2022/Authentication/AuthenticationFlowService.cs
andrewrobinsonhodges-snyk: Fixed

Snyk.VisualStudio.Extension.2022/Settings/SnykSettings.cs
andrewrobinsonhodges-snyk: Done

Snyk.VisualStudio.Extension.2022/Settings/HtmlSettingsControl.xaml.cs
```diff
@@ -0,0 +1,389 @@
...
+                SettingsBrowser,
+                dispatcher,
+                userDataFolder,
+                additionalInitScripts: new[]
+                {
+                    ExecuteCommandBridge.BuildClientScript(),
+                },
+                enableDeveloperTools: DeveloperToolsEnabled);
+
+            SettingsBrowser.NavigationCompleted += SettingsBrowser_OnNavigationCompleted;
```
bastiandoetsch, line 133: [Should Fix] `NavigationCompleted` is subscribed here but never unsubscribed in `Dispose()` — combined with the `IsVisibleChanged` handler (line 145), both leak and can fire on a disposed `host`, throwing `ObjectDisposedException` inside a `FireAndForget` that silently swallows it.

Fix:
```csharp
public void Dispose()
{
    if (_disposed) return;
    _disposed = true;
    SettingsBrowser.NavigationCompleted -= SettingsBrowser_OnNavigationCompleted;
    IsVisibleChanged -= Control_IsVisibleChanged;
    host?.Dispose();
}
```

— AI review

Snyk.VisualStudio.Extension.2022/UI/Html/HtmlSettingsScriptingBridge.cs
```diff
@@ -158,95 +189,263 @@ await ExecuteCommandBridge.DispatchAsync(
...
+            // without calling __saveIdeConfig__, so SaveCompletion would otherwise sit unsignalled
+            // until HtmlSettingsControl.SaveAsync's 5s timeout — leaving the Apply button hung and
+            // then surfacing a misleading "check the log" error. A non-"success" status lets
+            // OnApply report the failure immediately and keep the user on the page to fix it.
+            // The success path is still driven by __saveIdeConfig__ → TrySetResult(true);
+            // TrySetResult is first-wins, so a later "success" signal here is a harmless no-op.
+            if (!string.IsNullOrEmpty(status) &&
+                !string.Equals(status, "success", StringComparison.OrdinalIgnoreCase))
+            {
+                saveCompletionTcs.TrySetResult(false);
```
bastiandoetsch: [Should Fix] `saveCompletionTcs.TrySetResult(false)` reads the field directly without snapshotting it first, unlike `__saveIdeConfig__` (line 95) which snapshots `var tcs = saveCompletionTcs` for exactly this reason.

If a rapid double-Apply swaps `saveCompletionTcs` for a new instance between the JS `getAndSaveIdeConfig` invocation and this callback, `TrySetResult(false)` signals the wrong TCS — the second Apply's TCS gets a false failure signal while the first Apply's TCS times out after 5 seconds.

Fix (one line, mirrors the pattern already used in `__saveIdeConfig__`):
```csharp
public void __ideSaveAttemptFinished__(string status)
{
    var tcs = saveCompletionTcs;  // snapshot, same as __saveIdeConfig__ line 95
    Logger.Information("Save attempt finished with status: {Status}", status);
    if (!string.IsNullOrEmpty(status) &&
        !string.Equals(status, "success", StringComparison.OrdinalIgnoreCase))
    {
        tcs.TrySetResult(false);
    }
}
```

— AI review

bastiandoetsch:
**[Should Fix]** `Authenticate()` in `AuthenticationFlowService` has no re-entrancy guard.

If called twice before the first flow completes (e.g., user double-clicks "Authenticate"), two concurrent `RunAsync(...).FireAndForget()` lambdas both call `InvokeLogout` + `InvokeLogin` on the language server at the same time, and `AuthDialog.ArmForShow()` resets the dialog guard for the second call — so the auth dialog may appear twice or the two LS login sequences race.

`Snyk.VisualStudio.Extension.2022/Authentication/AuthenticationFlowService.cs`, `Authenticate()` method.

Fix: add an `_authInProgress` flag (or use `Interlocked`) to guard re-entrant calls:
```csharp
private volatile bool _authInProgress;

public void Authenticate()
{
    if (_authInProgress) return;
    _authInProgress = true;
    try
    {
        // ... existing body ...
    }
    finally
    {
        _authInProgress = false;
    }
}
```

— AI review

bastiandoetsch:
> ℹ️ These are non-blocking suggestions for your consideration.

**[Suggestion · non-blocking]** `AuthenticationFlowService.Authenticate()` uses `SnykVSPackage.Instance.DisposalToken` directly (lines 76-78) instead of going through the injected `ISnykServiceProvider`. Since the service is otherwise constructor-injected for testability, exposing `DisposalToken` on `ISnykServiceProvider` would complete the testability seam.

**[Suggestion · non-blocking]** `HtmlSettingsScriptingBridge` settings are serialized to JavaScript but no test covers the full property round-trip. A method rename or missing `[JsonProperty]` attribute on `IdeConfigData` would silently break settings with no failing test.

**[Suggestion · non-blocking]** `__saveIdeConfig__`/`__ideSaveAttemptFinished__` both use `Dispatcher.PushFrame` for the outer sync wait. The `saveInProgress` guard is sound for the re-entrant Apply case, but `PushFrame` also allows arbitrary VS events (tool-window visibility, timer callbacks) to fire while blocked. This is an accepted tradeoff for bridging the sync VS contract with async WebView2, but worth noting for future reviewers.

— AI review

Snyk.VisualStudio.Extension.2022/UI/Html/HtmlSettingsScriptingBridge.cs
andrewrobinsonhodges-snyk: Fixed

Snyk.VisualStudio.Extension.2022/Settings/HtmlSettingsControl.xaml.cs
andrewrobinsonhodges-snyk: Fixed

bastiandoetsch (approved):

PR #496, opened 2026-06-05
feat: add Secrets product to scan/results handling infrastructure [IDE-1841]

nick-y-snyk:
### Description

Adds the Secrets product to the VS 2022 scan and results handling infrastructure so Secrets findings flow through the same notification → cache → fire-event pipeline as OSS, Code, and IaC.

Stacked on top of #493 (HTML settings by default / WebView2 migration). UI tree nodes (root node, file/issue nodes, toolwindow subscriptions) are out of scope — they are a separate ticket.

**Changes:**
- `Product.cs` — `Product.Secrets = "secrets"` constant
- `SnykLanguageClientCustomTarget.cs` — secrets cache dict; `"Snyk Secrets"→"secrets"` source mapping; `OnPublishDiagnostics316` routing; `OnSnykScan` arm; `ProcessSecretsScanAsync`; `GetSecretsDictionary()` test accessor
- `LsAnalysisResult.cs` — `GetScoreComponent()` handles `Product.Secrets` (uses `RiskScore`, same as OSS/IaC per snyk-ls `SecretsIssueData`)
- `FeaturesSettings.cs` — `SecretsEnabled` property
- `SnykTasksService.cs` / `ISnykTasksService.cs` — `IsSecretsScanning` flag; 5 events (`SecretsScanningStarted/Update/Finished/Error/Disabled`); fire-methods; `ScanAsync` disabled guard; `IsTaskRunning`/`CancelTasks` include Secrets; `GetFeaturesSettingsAsync` maps `options.SecretsEnabled`

Wire strings sourced from snyk-ls: `$/snyk.scan` `product = "secrets"`, publishDiagnostics `source = "Snyk Secrets"`.

Jira: [IDE-1841](https://snyksec.atlassian.net/browse/IDE-1841)

### Checklist

- [ ] Linted
- [ ] CHANGELOG.md updated

### Screenshots / GIFs

N/A — infra only, no UI changes.

[IDE-1841]: https://snyksec.atlassian.net/browse/IDE-1841?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


bastiandoetsch:
Full verification pass (semantic, adversarial, security, code-review). The Secrets product is wired consistently across the producer surface — enum, feature-flag gating, diagnostics dictionary (add + clear paths), `OnSnykScan` routing, `LspSourceToProduct` mapping, and `Priority()` risk-score grouping — with no missing cases, and it's well-tested (incl. the `[InlineData(Product.Secrets, 400, 999, 3_000_400)]` priority case). The 5 new scanning events have no subscriber yet, but that's explicitly out of scope per the PR (UI tree nodes / toolwindow subscriptions are a separate ticket), so it's intentional staging, not a defect. Security scan: clean — the Secrets paths add no logging and never serialize finding content.

Worth folding in (ideally before the UI PR builds on this), none blocking for infra-only scope:
- [Should Fix] Secrets events use `SnykOssScanEventArgs` and `OnSecretsError` constructs `new SnykOssScanEventArgs(null, presentableError)`, diverging from IaC (a Code-engine product) which uses `SnykCodeScanEventArgs`. The `null` FeaturesSettings is a latent NRE if a future UI handler dereferences `e.FeaturesSettings`. Pick the event-arg type deliberately (likely `SnykCodeScanEventArgs`, matching IaC) and avoid passing null — cheaper to fix now than after a handler is written against the wrong type.
- [Suggestion] The success-path test asserts `FireSecretsScanningUpdateEvent` was called with an empty dictionary; add one test that publishes a Secrets diagnostic via `OnPublishDiagnostics316` then runs `OnSnykScan` and asserts the issue reaches the update event (proves the dictionary→event flow end-to-end).
- [Suggestion] `OnSecretsError` on `ISnykTasksService` is a fire-method wearing an `On*` (handler) name; `FireSecretsErrorEvent` would be clearer.
- [Note] `IsTaskRunning()`/`CancelTasks` now also include `IsIacScanning` (the base omitted it) — a correct latent IaC fix, but an unannounced behaviour change riding along in a Secrets PR; worth a line in the PR description.
- [Suggestion] `Product.cs` has no trailing newline; CHANGELOG checkbox unchecked.

— AI review

bastiandoetsch (approved):
Reviewed with the full verification pass. No correctness or security defects in the producer infrastructure — Secrets is threaded consistently through the enum, settings, diagnostics dictionary, scan routing, and risk-score scoring, mirroring the proven OSS/IaC patterns, and the tests cover the add/clear/in-progress/error/success paths. The absent UI consumer is intentional (separate ticket). Approving. One should-fix worth addressing before the UI PR depends on it: the `SnykOssScanEventArgs`/null-FeaturesSettings divergence from the IaC event-arg type (details in the comment).

— AI review

Snyk.VisualStudio.Extension.2022/Service/ISnykTasksService.cs
```diff
@@ -91,6 +92,31 @@ public interface ISnykTasksService
     /// </summary>
     event EventHandler<SnykCodeScanEventArgs> SnykCodeDisabled;
 
+    /// <summary>
+    /// Secrets scanning started event handler.
+    /// </summary>
+    event EventHandler<SnykOssScanEventArgs> SecretsScanningStarted;
```
bastiandoetsch, line 98: [Should Fix] All five Secrets events use `SnykOssScanEventArgs`, but Code and IaC events use `SnykCodeScanEventArgs` — the two types carry different cross-product state fields (`IacEnabled`, `OssScanRunning`, etc.). A future UI subscriber listening to `SecretsScanningStarted` will receive an OSS-shaped payload, which is misleading and may cause incorrect runtime behaviour.

Fix: Introduce a `SnykSecretsScanEventArgs` type following the IaC pattern, or at minimum add a code comment explicitly justifying why the OSS event args shape is correct for Secrets. The interface contract should be deliberate, not inherited by default.

— AI review

Snyk.VisualStudio.Extension.2022/Language/SnykLanguageClientCustomTarget.cs
```diff
@@ -77,6 +79,9 @@ public async Task OnPublishDiagnostics316(JToken arg)
                 case "iac":
                     snykIaCIssueDictionary.AddOrUpdate(parsedUri.UncAwareAbsolutePath(), issueList, (_, _) => issueList);
                     break;
+                case "secrets":
+                    snykSecretsIssueDictionary.AddOrUpdate(parsedUri.UncAwareAbsolutePath(), issueList, (_, _) => issueList);
+                    break;
                 default:
```
bastiandoetsch, line 85: [Should Fix] `LspSourceToProduct` returns an empty string `""` for any unrecognised source, but the `switch` in `OnPublishDiagnostics316` directly below this addition throws `InvalidProductTypeException` on its `default` branch. If the Language Server ever sends an unknown or misspelled source (e.g. `"Snyk Container"` for a future product, or a casing change), `LspSourceToProduct` silently returns `""`, the switch hits `default`, and the exception propagates inside the JsonRpc message pump — potentially breaking the LSP channel.

Fix: Handle the empty/unknown case explicitly. Either add a `case "": return;` guard before the switch, or make the switch default a graceful no-op (`default: Logger.Warning("Unknown product source: {source}"); return;`) instead of throwing.

— AI review

Snyk.VisualStudio.Extension.2022/Service/ISnykTasksService.cs
```diff
@@ -211,6 +237,14 @@ public interface ISnykTasksService
 
     void FireIacScanningUpdateEvent(IDictionary<string, IEnumerable<Issue>> analysisResult);
 
+    void FireSecretsScanningStartedEvent();
```
bastiandoetsch, line 240: [Suggestion] `FireSecretsScanningStartedEvent()` takes no parameters, while `FireIacScanningStartedEvent(FeaturesSettings)` and `FireSnykCodeScanningStartedEvent(FeaturesSettings)` both receive a `FeaturesSettings` argument. A future UI subscriber to `SecretsScanningStarted` will have no way to read `SecretsEnabled` or other feature flags from the started event — it would need to fetch settings separately.

Note: Semantic analysis confirmed the OSS product follows the same no-argument pattern, so this is a choice between aligning with OSS or with IaC/Code. If Secrets UI behaves more like OSS, no change is needed — but a comment documenting the intentional choice would prevent confusion.

Fix (if aligning with IaC/Code): `void FireSecretsScanningStartedEvent(FeaturesSettings featuresSettings);`

— AI review

Snyk.VisualStudio.Extension.2022/Service/SnykTasksService.cs
```diff
@@ -654,6 +687,47 @@ public void FireIacScanningFinishedEvent()
...
+            this.IsSecretsScanning = true;
+            this.SecretsScanningStarted?.Invoke(this, new SnykOssScanEventArgs());
+        }
+
+        /// <summary>
+        /// Fire Secrets scanning update event.
+        /// </summary>
+        public void FireSecretsScanningUpdateEvent(IDictionary<string, IEnumerable<Issue>> scanResult)
+        {
+            this.IsSecretsScanning = true;
```
bastiandoetsch, line 704: [Suggestion] `FireSecretsScanningUpdateEvent` sets `IsSecretsScanning = true` (line 704), but `FireSecretsScanningFinishedEvent` sets it back to `false` immediately after (no `await` between them). The flag flips true→false within the same call stack, so any observer on another thread will always see `false` by the time they read it.

Fix: Remove `IsSecretsScanning = true` from `FireSecretsScanningUpdateEvent`. Set `IsSecretsScanning = true` only in `FireSecretsScanningStartedEvent`, where it semantically belongs — the scan starts, not when an update fires mid-completion.

— AI review

Snyk.VisualStudio.Extension.2022/Service/SnykTasksService.cs
```diff
@@ -193,7 +219,7 @@ public static async Task InitializeAsync(ISnykServiceProvider serviceProvider)
         /// Check is Scan running (oss or snykcode) or CLI download.
         /// </summary>
         /// <returns>True if Oss or SnykCode scan running.</returns>
-        public bool IsTaskRunning() => this.IsOssScanning || this.IsSnykCodeScanning || this.isCliDownloading;
+        public bool IsTaskRunning() => this.IsOssScanning || this.IsSnykCodeScanning || this.IsIacScanning || this.IsSecretsScanning || this.isCliDownloading;
```
bastiandoetsch, line 222: [Suggestion] This correctly adds `IsIacScanning` to `IsTaskRunning()` — it was missing before, meaning a concurrent scan-trigger during an IaC-only scan would not have been blocked. The fix is right, but there is no test for this path.

Fix: Add a unit test asserting that `IsTaskRunning()` returns `true` when only `IsIacScanning` is `true` (and all others are `false`). Similarly for `IsSecretsScanning`.

— AI review

bastiandoetsch:
> ℹ️ This finding is in a file not changed by this PR. It is non-blocking for this review — shared for awareness.

[Should Fix · non-blocking] `TestUtils.SetupOptionsMock` does not configure `SecretsEnabled` on the shared mock. Any existing test that exercises `ScanAsync` now has a newly-reachable code path (`!selectedFeatures.SecretsEnabled → FireSecretsScanningDisabledEvent`) that fires when `SecretsEnabled` defaults to `false` on the mock. Existing event-count assertions may silently pass with incorrect state or break.

Fix: Add `optionsMock.SetupGet(o => o.SecretsEnabled).Returns(true)` to `TestUtils.SetupOptionsMock`, matching the established defaults for other products.

— AI review

bastiandoetsch:
> ℹ️ This finding is in files not changed by this PR. It is non-blocking for this review — shared for awareness.

[Suggestion · non-blocking] Two gaps in the UI wiring that will need to be closed in the follow-up ticket:

1. **`SnykToolWindowControl.xaml.cs:209`** — `SelectedItemInTree` switch has no case for `Product.Secrets`. Deep-link navigation from the Language Server (OnShowDocument with `product='Snyk Secrets'`) will be silently ignored until this is added.

2. **`SnykToolWindowControl.xaml.cs`** — None of the five Secrets scanning events (`SecretsScanningStarted/Update/Finished/Error/Disabled`) are subscribed to in `SnykToolWindowControl` or any other UI class, so Secrets results will not appear in the tree until the follow-up ticket wires them.

Both are expected given this PR's stated scope (infrastructure only, UI deferred). Noting here so they appear in the review trail before the next PR.

— AI review

bastiandoetsch:
> ℹ️ This finding is in dependencies not changed by this PR. It is non-blocking for this review — shared for awareness.

[Should Fix · non-blocking] Three pre-existing SCA vulnerabilities in transitive dependencies pulled in by the VS SDK test framework. None are introduced by this PR and none are directly upgradable (all locked by `Microsoft.VisualStudio.Shell.15.0`):

- `Microsoft.IO.Redist@6.0.0` — CVE-2024-38081 (Link Following)
- `System.Text.Json@6.0.0` — CVE-2024-43485 (Inefficient Algorithmic Complexity)
- `System.Text.Json@7.0.0` — CVE-2024-30105 (Denial of Service)

Fix: Track these with the VS SDK version bump whenever a newer `Microsoft.VisualStudio.Shell.15.0` is available that resolves the transitive chain.

— AI review

PR #506, opened 2026-06-17
feat: per-folder reset overrides [IDE-1945]

nick-y-snyk:
### Description

IDE-side support for the per-folder **Reset overrides** button (the button + dialog are served by snyk-ls). See snyk/snyk-ls#1344 for the LS side and the shared `FOLDER_RESET_FIELDS` list this mirrors.

A folder field sent by the dialog as JSON `null` means "reset this override" → the LS `Unset`s the `user:folder:` override (fallback to org/LDX/default). Native nullable types cannot distinguish present-`null` from absent, so the plugin re-parses the raw JSON tree and, for the reset-key whitelist, forwards `{value:null, changed:true}`.

This PR extends that whitelist to the 3 fields snyk-ls recently added so reset clears them too: `additional_parameters`, `additional_environment`, `scan_command_config`. See the commit body for plugin-specific details (naming fixes, stored-override clearing).

### Checklist

- [ ] README updated, if user-facing


Snyk.VisualStudio.Extension.2022/UI/Html/HtmlSettingsScriptingBridge.cs
```diff
@@ -608,14 +609,75 @@ private void ApplyMiscellaneousSettings(IdeConfigData config)
...
+                    if (property.Name == "folderPath")
+                        continue;
+                    if (property.Value.Type != JTokenType.Null)
+                        continue;
+
+                    // Flag the key so BuildFolderConfigs emits {value:null, changed:true} for it.
+                    // snyk-ls is authoritative: it acts on the keys it recognizes as folder-scoped
+                    // and ignores nulls on the rest, so forwarding every present-null is safe.
+                    existingConfig.ResetKeys ??= new HashSet<string>();
+                    existingConfig.ResetKeys.Add(property.Name);
```
basti-snyk: **Should Fix — `ResetKeys` is set here but never cleared, so the documented "transient + one-shot" contract isn't enforced.** `BuildFolderConfigs` (which consumes `ResetKeys`) runs on *every* `DidChangeConfiguration` and on every LS init/restart — but nothing in production ever clears `ResetKeys`. The only thing that drops it is an inbound `$/snyk.configuration` echo rebuilding the `FolderConfig` objects via `FolderConfigApplier.Apply`, and that push is fire-and-forget — not awaited, not ordered before the next save, and a no-op on a global-only echo.

Concrete failure: user resets `snyk_code_enabled` for a folder (correct), then re-enables it and saves again before an echo arrives. `SaveFolderConfigsAsync` sets `SnykCodeEnabled = true`, but the stale `ResetKeys` still contains the key, so `BuildFolderConfigs`'s reset loop overwrites the just-set value with `{value:null, changed:true}` — silently un-setting the user's new override. Also: because the stored typed value isn't cleared on reset, the object carries both a stale value and a stale ResetKey, so the reset re-fires on every subsequent send until an echo rebuilds it.

Fix: clear `ResetKeys` immediately after `BuildFolderConfigs` emits them (or compute resets into a transient structure not stored on the persisted `FolderConfig`). Add a test that calls the build path twice after a reset and asserts the reset is emitted only once.

— AI review

nick-y-snyk: Fixed in 5565a7d. Took the root-cause route: the VS plugin was persisting folder configs to disk and re-sending them on init/restart — the source of the stale-state class. Now in-memory only, matching vscode/eclipse (LS is authoritative and re-pushes via `$/snyk.configuration`). On top of that, `FolderConfig.Set`/`SetString` now drop the key from `ResetKeys` so a re-set wins over a pending reset, killing the same-session clobber you described. The "Reset Overrides Not Persisted" disk variant is moot now too.

Snyk.VisualStudio.Extension.2022/Language/CustomInitializationOptions.cs
```diff
@@ -16,42 +17,64 @@ public CamelCasePreserveDictionaryKeysNamingStrategy() : base(processDictionaryK
...
+        public HashSet<string> ResetKeys { get; set; }
+
+        // ----- Typed accessors over the opaque map (keyed by PflagKeys.*) -----
+        // These let the dialog and the handful of IDE-side readers stay readable without
+        // re-introducing typed fields. Values are stored as ConfigSetting.Of(...) so the round-trip
+        // back to the LS carries Changed=true.
+
+        public string GetString(string key) => GetValueToken(key)?.Value<string>();
+
+        public List<string> GetStringList(string key) => GetValueToken(key)?.ToObject<List<string>>();
```
basti-snyk, line 52: **Should Fix — `GetStringList`/`GetString` now throw at the consumer instead of degrading to null.**

The old per-key `ToFolderConfig` extraction wrapped each value in try/catch and skipped/logged on failure. The opaque map is now copied verbatim, and `GetStringList` does `GetValueToken(key)?.ToObject<List<string>>()` with no guard. `GetLsDebugLevelAsync` (`SnykLanguageClient`) calls `GetStringList(AdditionalParameters)` over every folder config on LS (re)start, outside a try/catch. Global `additional_parameters` is sent as a space-joined **string** while folder values are arrays — if a folder's `additional_parameters` is ever stored/echoed as a scalar or string, `ToObject<List<string>>()` throws `JsonSerializationException` out of the LINQ pipeline and aborts debug-level detection / LS startup, rather than falling back to `info`. Consider try/catch-and-degrade here (restoring the old defensiveness), or asserting the value shape.

— AI review

Snyk.VisualStudio.Extension.2022/BranchSelectorDialogWindow.xaml.cs
```diff
@@ -49,8 +49,8 @@ private void OkButton_OnClick(object sender, RoutedEventArgs e)
             {
                 return;
             }
-            FolderConfig.BaseBranch = CbBranchList.SelectedItem.ToString();
-            FolderConfig.ReferenceFolderPath = SelectedFolderPath.Text;
+            FolderConfig.SetString(PflagKeys.BaseBranch, CbBranchList.SelectedItem.ToString());
```
basti-snyk: **Should Fix (touched line) — OK handler dereferences `CbBranchList.SelectedItem` with no null check.**

The only guard in `OkButton_OnClick` is `if (FolderConfig == null) return;`. If no branch is selected (e.g. `base_branch` unset so `SelectedItem` was initialized to `GetString(BaseBranch)` = null, or `LocalBranches` is empty), `SelectedItem.ToString()` here throws `NullReferenceException` and the dialog crashes, losing the user's reference-folder selection on the same OK click. The pre-existing line had the same shape, but since it's being touched anyway a `?.ToString()` / explicit null guard would close it.

— AI review

Snyk.VisualStudio.Extension.2022/BranchSelectorDialogWindow.xaml.cs
```diff
@@ -49,8 +49,8 @@ private void OkButton_OnClick(object sender, RoutedEventArgs e)
             {
                 return;
             }
-            FolderConfig.BaseBranch = CbBranchList.SelectedItem.ToString();
-            FolderConfig.ReferenceFolderPath = SelectedFolderPath.Text;
+            FolderConfig.SetString(PflagKeys.BaseBranch, CbBranchList.SelectedItem?.ToString());
```
basti-snyk, line 52: **Should Fix — the Branch Selector can no longer *reset* `base_branch` / `reference_folder`; it can only set or (for base branch) silently drop them.** With the move to the opaque map, `SetString` has reset-incompatible null semantics: `SetString(key, null)` *removes* the key (no `{value:null, changed:true}` reset is emitted), and `SetString(key, "")` stores a real empty-string override.

Two concrete consequences here:
- **Base branch:** when no branch is selected, `CbBranchList.SelectedItem?.ToString()` is `null`, so `SetString(BaseBranch, null)` removes the key. The LS receives no change and retains the stale base branch instead of having it reset.
- **Reference folder:** clearing the textbox makes `SelectedFolderPath.Text == ""`, so `SetString(ReferenceFolder, "")` sends `{value:"", changed:true}` — setting an empty override rather than unsetting it (the delta-scan baseline becomes an empty path instead of reverting to default).

The form-driven path in `HtmlSettingsScriptingBridge` correctly uses `Set(key, null)` for resets; this dialog should do the same for the clear/empty cases (e.g. `Set(PflagKeys.BaseBranch, null)` when nothing is selected, and treat empty reference-folder text as a reset). A test driving OK with no selection / cleared text and asserting a `{value:null, changed:true}` reset would lock this in.

> This is a re-review at head 3db9695. The earlier SelectedItem null-deref is resolved; this is the distinct reset-semantics issue on the same lines.

— AI review

basti-snyk (approved):

PR #507, opened 2026-06-19
feat(settings): wire global additional_parameters and close additional_environment HTML-save hole [IDE-2110]

nick-y-snyk:
## Summary

- **`additional_environment` HTML-save hole (Part A):** `IdeConfigData` had no top-level `AdditionalEnv` field, so the HTML form's Project Defaults save silently dropped the value. Adds `[JsonProperty("additional_environment")]` to `IdeConfigData` and applies it in `ApplyMiscellaneousSettings`. Extends `SnapshotOptionsForRollback` to capture/restore `AdditionalParameters`.

- **`additional_parameters` global (Part B, end-to-end):** previously missing entirely at global scope. Adds:
  - `List<string> AdditionalParameters` to `IPersistableOption`, `SnykOptions`, `SnykSettings` (default `new List<string>()`), `SnykOptionsManager` load+save
  - `BuildSettingsMap` entry: `[PflagKeys.AdditionalParameters] = ConfigSetting.Of(options.AdditionalParameters ?? new List<string>())`
  - `GlobalSettingsApplier` case: `options.AdditionalParameters = val.ToObject<List<string>>()`
  - Top-level `[JsonProperty("additional_parameters")]` on `IdeConfigData`; applied in `ApplyMiscellaneousSettings`

Both keys use the same string as the per-folder `FolderConfigData` fields — the LS disambiguates by map position (top-level `settings` vs `folderConfigs[].settings`).

Fixes the `BuildSettingsMap_AdditionalParameters_NotInGlobalMap` test (flipped to `_InGlobalMap`); adds `AdditionalEnvironment` in-global-map assertion; adds both keys to the `IdeConfigContract` authoritative payload.

Completes IDE-side wiring for snyk-ls PR #1340 (IDE-2110).

## Test plan

- [ ] `dotnet build` green (Windows required)
- [ ] `dotnet test` — `LsSettingsV25Tests`, `IdeConfigContractTests`, `HtmlSettingsScriptingBridgeTest` pass
- [ ] Manual: open Settings → Project Defaults → Advanced; set both fields; save; reopen — fields repopulate
- [ ] OSS scan picks up global env vars and CLI params

### DIFF
--- a/Snyk.VisualStudio.Extension.2022/Language/SnykLanguageClient.cs
+++ b/Snyk.VisualStudio.Extension.2022/Language/SnykLanguageClient.cs
@@ -224,16 +224,18 @@ namespace Snyk.VisualStudio.Extension.Language
         private Task<string> GetLsDebugLevelAsync()
         {
             var serviceProvider = SnykVSPackage.ServiceProvider;
-            var folderConfigs = serviceProvider?.Options?.FolderConfigs;
-            if (folderConfigs == null)
+            var options = serviceProvider?.Options;
+            if (options == null)
                 return Task.FromResult("info");
 
-            // Enable debug logging for the whole LS process if -d/--debug is set on ANY folder's
-            // additional parameters, not just the first — workspaces can have multiple folders.
-            var anyDebug = folderConfigs
+            // Enable debug logging for the whole LS process if -d/--debug is set in global
+            // additional parameters OR in ANY folder's additional parameters.
+            var globalParams = options.AdditionalParameters ?? Enumerable.Empty<string>();
+            var folderParams = (options.FolderConfigs ?? Enumerable.Empty<FolderConfig>())
                 .Where(fc => fc?.AdditionalParameters != null)
-                .SelectMany(fc => fc.AdditionalParameters)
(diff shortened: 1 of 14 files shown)


rrama:
AI found bug (unchecked by me):

> **GetLsDebugLevelAsync does not check global AdditionalParameters for --debug flag**
> 
> `GetLsDebugLevelAsync` at `Snyk.VisualStudio.Extension.2022/Language/SnykLanguageClient.cs:224-239` only checks per-folder `FolderConfig.AdditionalParameters` for `-d`/`--debug` flags. This PR promotes `AdditionalParameters` from a folder-only setting to also being a global setting (via `ISnykOptions.AdditionalParameters`), but `GetLsDebugLevelAsync` was not updated to also inspect `serviceProvider.Options.AdditionalParameters`. If a user sets `--debug` in the global Project Defaults additional parameters (and no folder config also has it), the LS will start with `"info"` level instead of `"debug"`.

rrama (approved):
Other than the AI found bug for `--debug`, rest seems fine to me. It may be easier to to the `--debug` stuff in another follow-on PR.

rrama (approved):

PR #525, opened 2026-07-23
chore: sync settings-fallback.html from snyk-ls

team-ide-user:
Automatic sync of `settings-fallback.html` triggered by [snyk/snyk-ls@920eb24](https://github.com/snyk/snyk-ls/commit/920eb24fdb6677d8767aa86ca6153242ca80f535).

This PR was opened by the [distribute-fallback-html](https://github.com/snyk/snyk-ls/blob/main/.github/workflows/distribute-fallback-html.yaml) workflow in snyk-ls.

## What changed

`settings-fallback.html` is the settings page displayed before the Language Server binary is available. It is maintained in [snyk/snyk-ls](https://github.com/snyk/snyk-ls) and must be kept in sync with the copy in each IDE plugin. This repo's `resource-check` CI workflow verifies that the local copy matches the snyk-ls main branch; this PR restores that sync.

## Merge instructions

Review and merge when ready. No manual testing is required beyond confirming that `resource-check` passes on this PR.

### DIFF
--- a/Snyk.VisualStudio.Extension.2022/Resources/settings-fallback.html
+++ b/Snyk.VisualStudio.Extension.2022/Resources/settings-fallback.html
@@ -59,6 +59,14 @@
       --border-color: var(--vscode-panel-border, #454545);
       --focus-border: var(--vscode-focusBorder, #007acc);
 
+      /* Buttons */
+      --button-background: var(--vscode-button-background, #0e639c);
+      --button-foreground: var(--vscode-button-foreground, #ffffff);
+      --button-hover-background: var(--vscode-button-hoverBackground, #1177bb);
+      --button-secondary-background: var(--vscode-button-secondaryBackground, #3a3d41);
+      --button-secondary-foreground: var(--vscode-button-secondaryForeground, #ffffff);
+      --button-secondary-hover-background: var(--vscode-button-secondaryHoverBackground, #45494e);
+
       /* Other */
       --border-radius: 4px;
       --border-radius-small: 2px;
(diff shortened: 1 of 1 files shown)


rrama (approved):

rrama:
Checked all three findings from the review bot ([comment](https://github.com/snyk/snyk-visual-studio-plugin/pull/525#issuecomment-5060499802), [comment](https://github.com/snyk/snyk-visual-studio-plugin/pull/525#issuecomment-5071426711)) against the actual bridge/theming code — all three are false positives, traced below.

**"Incompatible JS Bridge" [major]** — claims no generic `__ideExecuteCommand__` dispatcher exists, so the logout button silently fails. It doesn't check the settings window's own bridge, only the per-issue-type providers (`CodeHtmlProvider`/`SecretsHtmlProvider`, which use one-off named bindings like `window.OpenFileInEditor`). The settings window has its own purpose-built bridge, already present at the reviewed commit:
- `HtmlSettingsWindow.xaml.cs` registers `__ideExecuteCommand__` as a host object and injects `ExecuteCommandBridge.BuildClientScript()` as an init script
- `HtmlSettingsScriptingBridge.__ideExecuteCommand__` forwards to `ExecuteCommandBridge.DispatchAsync`
- `ExecuteCommandBridge.AllowedCommands` explicitly allowlists `snyk.logout`

Fully wired end-to-end; the button does not silently fail.

**"Misleading Feedback" [minor]** — claims `done()` shows 'Signed out.' even if logout fails. Traced the whole path: the LS command (`domain/ide/command/logout.go`, snyk-ls) always returns `(nil, nil)` — it can't signal failure through its result. The only failure mode is a C# exception in `ExecuteCommandBridge.DispatchAsync`, which is caught and logged but does **not** invoke the JS callback at all in that branch. So `done()` only ever fires on success today — there's no path where it fires with a failure and shows a false 'Signed out.'

**"Theme Variable Mismatch" [minor]** — claims the new `--vscode-button-*` variables aren't recognised by the theming code, citing `CodeHtmlProvider.cs`'s hardcoded `.Replace("var(--button-background-color)", ...)` as "the standard." But `settings-fallback.html` doesn't go through `CodeHtmlProvider` — it's loaded via `HtmlResourceLoader.LoadFallbackHtml`, which calls the generic `BaseHtmlProvider.ReplaceCssVariables`. That method's `ideVarOverridesMap` already defines `vscode-button-background`, `vscode-button-hoverBackground`, `vscode-button-secondaryBackground`, `vscode-button-secondaryForeground`, and `vscode-button-secondaryHoverBackground` — exactly the variables this PR introduces. They'll theme correctly against the active VS color, not fall back to the hardcoded hex.

No changes needed on account of any of these.

PR #560, opened 2026-09-10
fix: remove the remaining eager static Serilog logger fields [IDE-2558]

basti-snyk:
### Description

Stacked on #559. Review that one first.

#559 fixed one crash site. This closes the class.

`SnykVSPackage` had `private static readonly ILogger Logger = LogManager.ForContext<T>();`. A static field initializer runs inside the type's `.cctor`, so the first touch of the type forced a Serilog load before any of the extension's own error handling could run. When that load failed, Visual Studio showed "The SnykVSPackage package was not loaded correctly" and the extension never started.

Two reviewers on #559 pointed out that 26 other classes still declared the same field, and that the repo had already settled how to avoid it: `SettingsLocationMigrator` carries an explicit "no static ILogger field here" comment, and three other classes follow it by calling `LogManager.ForContext<T>()` from a getter instead.

This applies that convention to the remaining 26. Each change is one line and no call sites move.

**Thirteen are a real crash risk**, because the host constructs them rather than Snyk code, so they can be first-touched outside any Snyk `try`/`catch` exactly as `SnykVSPackage` was:

- `SnykLanguageClient` (MEF `[Export(typeof(ILanguageClient))]`) and `SnykLanguageClientMiddleware`
- `HtmlSettingsDialogPage` (`[ProvideOptionPage]`)
- Six XAML `UserControl`s: `HtmlSettingsControl`, `TreeHtmlPanel`, `SnykToolWindowControl`, `HtmlDescriptionPanel`, `SummaryHtmlPanel`, `MessagePanel`
- Four types built inside those constructors: `SnykScriptManager`, `WebView2Host`, `WebView2NavigationPreparer`, `HtmlSettingsScriptingBridge`

**Thirteen are consistency only**, constructed from Snyk code that has already initialized: the `Service/` classes, `SnykVsThemeService`, `AuthenticationFlowService`, `SnykCli`, `SnykCliDownloader`, `SnykLanguageClientCustomTarget`, `GlobalSettingsApplier`, `WebView2MessageDispatcher`, `ExecuteCommandBridge`.

`LogManager.ForContext` already caches the underlying logger behind its own `Lazy<Logger>`, so the per-call cost is a wrapper allocation rather than logger construction. Hot paths were checked rather than assumed. `SnykLanguageClientMiddleware.HandleNotificationAsync` was the strongest candidate, but it only logs when the method is `window/logMessage`, so it is bounded by log-forwarding volume and not by every RPC. Nothing logs inside a tight loop, so no class was excepted on performance grounds.

The regression pin from #559 now scans every type in the extension assembly instead of just `SnykVSPackage`, so a reintroduction anywhere fails the build. `LogManager` is the one allow-listed type: it is the Serilog bridge itself, and its static Serilog fields are the guarded implementation everything else routes through.

Also carries one fix to #559's own code that was found after that PR was squashed: the assembly resolver traced every routine decline, and since it is subscribed to the process-wide `AssemblyResolve` event it saw every failed bind in Visual Studio, including frequent satellite-resource probes. `TRACE` is defined in Release, so that shipped. The traces that remain mark a successful resolve, a refused name, and caught exceptions, all of which are rare and worth reading.

### Checklist

- [x] Read and understood the [Code of Conduct](https://github.com/snyk/snyk-visual-studio-plugin/blob/main/CODE_OF_CONDUCT.md) and [Contributing Guidelines](https://github.com/snyk/snyk-visual-studio-plugin/blob/main/CONTRIBUTING.md).
- [ ] Linted
- [x] README.md updated, if user-facing

**On the unchecked boxes, honestly:** the regression pin was extended, but this work was done in a Linux environment with no MSBuild or VS SDK, so the suite was never compiled or run locally. CI is the first execution. No README change is needed, since nothing here is user-facing.

Manual verification still needs doing by someone with a Visual Studio instance: install the VSIX alongside another extension that bundles its own Serilog, open and dock the Snyk tool window, restart VS, and confirm no "package was not loaded correctly" dialog.

---

> [!NOTE]
> **Medium Risk**
> Serilog major-version bump plus process-wide binding-path registration affect extension startup and logging across the whole VS session; incorrect binding could still break load or clash with other extensions.
> 
> **Overview**
> This PR finishes **IDE-2558** work on two fronts: it stops **eager Serilog loads during type construction**, and it adds a **binding stopgap** so the extension can load when GitLab for Visual Studio is installed.
> 
> **Lazy loggers:** Every remaining `private static readonly ILogger` initializer becomes a **property getter** that calls `LogManager.ForContext<T>()`, matching the pattern already used on `SnykVSPackage`. That defers the first Serilog touch until something actually logs, instead of running it in each type's static constructor when VS/MEF constructs those types early.
> 
> **Load fix:** The project now **directly references Serilog 4.4.0** (and bumps related Serilog packages) so the bound identity sits **above** GitLab's process-wide redirect ceiling, and **`[ProvideBindingPath]`** is added on `SnykVSPackage` so the binder probes the extension install folder. **`ExtensionAssemblyResolver`** no longer traces every routine assembly-resolve decline (a Release hot path).
> 
> **Tests & docs:** Assembly-wide **`NoSerilogTypedStaticFieldTests`** replaces the package-only pin; **`SerilogAssemblyIdentityTests`** asserts the shipped Serilog version and reference align with the redirect strategy. The IDE-2558 plan doc is rewritten to document root cause, the stopgap, and rejected alternatives.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 7546f823a974b1996cb50eea4e9cbb70a3093ad9. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


rrama:
Need more work, see the Jira ticket for the result of my manual testing.

